PR #201, opened 2026-06-21
Fix #194: allow breaking the mirror to disable the room teleport

### LINKED ISSUE #194: Mirror can't be broken (mirror-room teleport never disabled)
The mirror in the mirror rooms can't be broken, so the teleport works forever.

**Original (`zork1/1actions.zil` MIRROR-MIRROR / MIRROR-ROOM):** mung/throw/attack breaks the mirror permanently (`MIRROR-MUNG`), disabling the touch-to-teleport and noting the destruction in the room description.

**C#:** `ZorkOne/Location/MirrorRoom*.cs` / `Mirror.cs` have no break path; the mirror is usable indefinitely.

**Fix:** allow breaking the mirror (mung/throw/attack), set a broken flag that disables the teleport and updates the description. Add tests.

_Found via the ZIL audit; minor._

arsindelve:
Fixes #194.

## Problem

The enormous mirror in the two mirror rooms could never be broken, so the touch-to-teleport between them worked forever.

## Original behavior (ground truth)

Verified against the original Infocom ZIL — `1actions.zil`, the `MIRROR-MIRROR` routine:

- `MUNG`/`THROW`/`ATTACK` sets the global `MIRROR-MUNG` flag and prints: *"You have broken the mirror. I hope you have a seven years' supply of good luck handy."*
- Attacking it again (when already `MIRROR-MUNG`) prints: *"Haven't you done enough damage already?"*
- `EXAMINE` prints *"The mirror is broken into many pieces."* when broken, else *"There is an ugly person staring back at you."*
- `RUB` is guarded by `<NOT ,MIRROR-MUNG>`, so a broken mirror no longer teleports.

The mirror's action only fires when the mirror itself is the direct object, so single-noun targeting matches the original (throwing a *held item* at it does not trigger it in the ZIL either).

## Fix

- Added a shared `IsBroken` flag to the `Mirror` item. A single `Mirror` instance is shared by both rooms, mirroring the original global `MIRROR-MUNG` flag — break it in one room and the teleport is disabled in both.
- Break verbs set the flag and print the canonical bad-luck message; a second attempt prints *"Haven't you done enough damage already?"*
- Once broken: rub/touch no longer teleports, examine reports the broken state, and the room description notes the shattered remains.
- Break-verb coverage reuses `Verbs.KillVerbs` / `Verbs.ThrowVerbs` (per the CLAUDE.md *"Add verb synonym → Model/Verbs.cs"* convention) plus the break-specific verbs the original recognized.
- The broken-mirror string lives in one place (`Mirror.BrokenDescription`).

## Tests (TDD)

`ZorkOne.Tests/Places/MirrorRoomTests.cs` (9 tests): teleport-before-break, breaking via `kick`/`throw`/`punch` (one from each verb source), the already-broken message, the disabled teleport after breaking (including across rooms via the shared instance), and the updated examine/room descriptions. All 9 new tests pass; full `ZorkOne.Tests` (1175) and `UnitTests` (820) suites remain green.


ZorkOne/Location/MirrorRoom.cs
```diff
@@ -21,11 +26,32 @@ public override async Task<InteractionResult> RespondToSimpleInteraction(SimpleI
         if (!action.MatchNoun(["mirror"]))
             return await base.RespondToSimpleInteraction(action, context, client, itemProcessorFactory);
 
+        var mirror = GetItem<Mirror>();
+
+        // Breaking the mirror (mung/throw/attack in the original ZIL MIRROR-MIRROR) permanently disables
+        // the teleport. Reuse the shared attack/throw synonym sets so the verb coverage stays consistent
+        // with the rest of the engine, plus the break-specific verbs the original recognized.
+        if (action.MatchVerb([..Verbs.KillVerbs, ..Verbs.ThrowVerbs, ..Verbs.BreakVerbs, "kick", "hit"]))
```
arsindelve: Extract this into a common private array, used twice

ZorkOne/Location/MirrorRoom.cs
```diff
@@ -21,11 +26,32 @@ public override async Task<InteractionResult> RespondToSimpleInteraction(SimpleI
         if (!action.MatchNoun(["mirror"]))
             return await base.RespondToSimpleInteraction(action, context, client, itemProcessorFactory);
 
+        var mirror = GetItem<Mirror>();
+
+        // Breaking the mirror (mung/throw/attack in the original ZIL MIRROR-MIRROR) permanently disables
```
arsindelve: Remove the reference to the ZIL source code in this comment

ZorkOne/Location/MirrorRoom.cs
```diff
@@ -21,11 +26,32 @@ public override async Task<InteractionResult> RespondToSimpleInteraction(SimpleI
...
+                "You have broken the mirror. I hope you have a seven years' supply of good luck handy. ");
+        }
+
         if (action.MatchVerb(["look", "examine", "peer"]))
-            return new PositiveInteractionResult("There is an ugly person staring back at you. ");
+            return new PositiveInteractionResult(mirror.IsBroken
+                ? Mirror.BrokenDescription
+                : "There is an ugly person staring back at you. ");
 
         if (action.MatchVerb(["rub", "touch", "feel", "press"]))
```
arsindelve: This feels like another candidate for verbs.cs plus some other common additionas.

ZorkOne/Location/MirrorRoom.cs
```diff
@@ -7,10 +7,21 @@ namespace ZorkOne.Location;
 
 internal abstract class MirrorRoom : LocationBase, IThiefMayVisit
 {
+    // Breaking the mirror permanently disables the touch-to-teleport. Reuse the shared attack/throw synonym
+    // sets so the verb coverage stays consistent with the rest of the engine, plus the break-specific verbs.
+    // Built once, not per call.
+    private static readonly string[] BreakMirrorVerbs =
+        [..Verbs.KillVerbs, ..Verbs.ThrowVerbs, ..Verbs.BreakVerbs, "kick", "hit"];
```
arsindelve: We can't have throw here without a second noun. Throw the mirror? Makes no sense,

ZorkOne/Location/MirrorRoom.cs
arsindelve: Extracted into a `private static readonly string[] BreakMirrorVerbs` (built once instead of re-allocated per call). Heads-up: it was actually referenced only once in the file — the thing used twice is the broken-mirror *string*, which is already centralized as `Mirror.BrokenDescription`. Named the array anyway since it's cleaner.

---

ZorkOne/Location/MirrorRoom.cs
arsindelve: Agreed — removed `ThrowVerbs` from the mirror break set. "throw mirror" now falls through (no break), and I updated the test to assert that. Throwing an object *at* the mirror would be a separate multi-noun interaction.

---

arsindelve: I think we need to keep throw here, and implement a multi noun resolver. Any object, thrown at the mirror, breaks the mirror and puts the item on the floor. But there must be some objects (garlic? brown sack?) which would not break the mirror. I dont want to overcomplicate this. Discuss.

arsindelve: Implemented (commit `469e4b9`), based on your answers to the two decisions:

- **`throw <X> at mirror`** now goes through a `RespondToMultiNounInteraction` override on `MirrorRoom` (matches `Verbs.ThrowVerbs` + mirror as the target noun; requires you to actually hold the item).
- **Default:** the item shatters the mirror (sets `IsBroken`, disabling the teleport) and lands on the floor.
- **Soft exceptions:** `Garlic`, `BrownSack`, and `Lunch` bounce off and land on the floor, mirror intact. Explicit list (not size-based — sizes don't cleanly separate soft from hard). Easy one-liner to add to: `thrownItem is Garlic or BrownSack or Lunch`.
- **Already broken:** "Haven't you done enough damage already?" and the item just drops.
- **Scope = break + drop only** — no secondary effects (the thrown item itself isn't damaged, light sources stay lit).

Single-noun `throw mirror` still correctly does nothing. Tests cover both the hard-object break+drop and the soft-object bounce+drop, routed through the real multi-noun parser path. One note for ground truth: the original ZIL had **no** soft-item exemption (anything thrown broke it), so the garlic/sack carve-out is a deliberate, nice-to-have divergence — flagging in case you'd rather match the original exactly.

---

ZorkOne/Location/MirrorRoom.cs
```diff
@@ -36,4 +65,37 @@ public override async Task<InteractionResult> RespondToSimpleInteraction(SimpleI
...
+            return new PositiveInteractionResult("You don't have that. ");
+
+        var mirror = GetItem<Mirror>();
+        var itemName = thrownItem.NounsForMatching.OrderByDescending(s => s.Length).First();
+
+        // The item leaves your hands and lands on the floor of this room, broken mirror or not.
+        ItemPlacedHere(thrownItem);
+
+        // Soft, lightweight items aren't going to shatter a wall-sized mirror.
+        if (thrownItem is Garlic or BrownSack or Lunch)
```
arsindelve: Audit every item in the game and make sure you haven't missed any "soft" items. Then decide on a place to put this code (ItemBase? Zork!Context) - it does not belong here. MirrorRoom should not decide what is soft or not - SRP.

arsindelve:
There are merge conflicts to resolve

ZorkOne/Location/MirrorRoom.cs
arsindelve: Done (commit `bdc7ed8`).

**SRP:** moved the decision onto the items. Added `IItem.IsSoft` — a `virtual bool` on `ItemBase` defaulting to `false` (same pattern as `Size`). Each item now declares its own softness, and `MirrorRoom` just reads `thrownItem.IsSoft` — no more type list in the room.

**Audit:** went through every takeable Zork I item. Flagged these 13 as soft: `Garlic`, `Lunch`, `BrownSack`, `Leaflet`, `Manual`, `Map`, `Guidebook`, `Matchbook`, `PileOfLeaves`, `Rope`, `ViscousMaterial`, `Candles`, `PileOfPlastic` (food / paper / cloth / leaves / rope / putty / wax / deflated plastic). Judgment calls worth flagging: I kept the **clockwork canary** (it's solid metal), **coal**, and **timber** as *hard* (they'd realistically shatter glass) — shout if you'd rather treat any of those as soft. Added a test that locks the audit and the default. All suites green.

Also resolved the merge conflicts (main had independently added `Verbs.ExamineVerbs` from #207 — kept both that and my `TouchVerbs`, and the mirror's examine now uses `ExamineVerbs`).

---

ZorkOne.Tests/Places/MirrorRoomTests.cs
```diff
@@ -0,0 +1,191 @@
...
+        // You can't "throw" a wall-sized mirror with no second noun; throwing something AT it is a
+        // separate (multi-noun) interaction. So a bare "throw mirror" must not break it.
+        var target = GetTarget();
+        target.Context.CurrentLocation = Repository.GetLocation<MirrorRoomSouth>();
+
+        var response = await target.GetResponse("throw mirror");
+        Console.WriteLine(response);
+
+        response.Should().NotContain("broken the mirror");
+        Repository.GetItem<Mirror>().IsBroken.Should().BeFalse();
```
arsindelve, line 56: Assert we get an AI response here.

ZorkOne/Location/MirrorRoom.cs
```diff
@@ -36,4 +65,38 @@ public override async Task<InteractionResult> RespondToSimpleInteraction(SimpleI
...
+
+    public override async Task<InteractionResult?> RespondToMultiNounInteraction(MultiNounIntent action,
+        IContext context)
+    {
+        // "throw <X> at mirror": anything you hurl at the mirror shatters it (and disables the teleport),
+        // except a few soft, lightweight items that just bounce off. Either way the item ends up on the floor.
+        if (!action.MatchVerb(Verbs.ThrowVerbs) || !action.MatchNounTwo(["mirror"]))
+            return await base.RespondToMultiNounInteraction(action, context);
+
+        var (hasItem, thrownItem) = context.HasMatchingNoun(action.NounOne);
```
arsindelve, line 77: Add a unit test for this

ZorkOne/Location/MirrorRoom.cs
```diff
@@ -36,4 +65,38 @@ public override async Task<InteractionResult> RespondToSimpleInteraction(SimpleI
...
+        if (!hasItem || thrownItem is null)
+            return new PositiveInteractionResult("You don't have that. ");
+
+        var mirror = GetItem<Mirror>();
+        var itemName = thrownItem.NounsForMatching.OrderByDescending(s => s.Length).First();
+
+        // The item leaves your hands and lands on the floor of this room, broken mirror or not.
+        ItemPlacedHere(thrownItem);
```
arsindelve, line 85: Add a unit test for this (the item is now on the floor) if you have not already

ZorkOne.Tests/Places/MirrorRoomTests.cs
arsindelve: Done — `ThrowMirror_DoesNotBreakIt_AndFallsBackToAi` now stubs the generation client and asserts the response is the AI `VerbHasNoEffect` narration ("Throwing the mirror accomplishes nothing."), confirming the fall-through path actually reaches the AI rather than silently returning nothing.

---

ZorkOne/Location/MirrorRoom.cs
arsindelve: Added `ThrowObjectYouDoNotHaveAtMirror_DoesNotBreakIt` — "throw sword at mirror" while not holding the sword returns "You don't have that." and leaves the mirror intact.

---

ZorkOne/Location/MirrorRoom.cs
arsindelve: Covered. The break case (`ThrowHardObjectAtMirror_BreaksIt_AndDropsItOnFloor`) and the bounce case (`ThrowSoftObjectAtMirror_BouncesOff_AndDropsItOnFloor`) both assert the item leaves inventory and is on the room floor. I also added `ThrowObjectAtAlreadyBrokenMirror_DropsItOnFloor` so the already-broken branch's drop is asserted too.

---

PR #229, opened 2026-06-21
Restore missing object synonyms from the original ZIL (#228)

### LINKED ISSUE #23: Put the scepter in the case when it's already in the case.
It noun-matches the first noun on the case because it contains an item with a noun match for scepter. Then it tried to put the case in the case.

### LINKED ISSUE #228: Restore missing object synonyms/adjectives from the original ZIL (parser vocabulary fidelity)
## Summary

The original Infocom games give each object a rich `(SYNONYM …)` and `(ADJECTIVE …)` vocabulary so the parser recognizes the many words a player might use. The C# port's `NounsForMatching` arrays are, for many objects, a **subset** of that vocabulary — so some perfectly reasonable references silently fail to resolve. This is the **same class of bug** as the already-fixed #23 (the sceptre couldn't be called "scepter"). #23 was found by accident; this issue is the systematic version, produced by mechanically diffing every ZIL `SYNONYM`/`ADJECTIVE` against `NounsForMatching` for both games.

None of the structural/behavior audits catch this — the objects exist and behave correctly; they just can't be *named* every way the original allowed.

## ⚠️ Important caveat (verify each before adding)

The diff **over-reports**, because some synonyms are handled contextually via per-verb matching rather than `NounsForMatching`. Confirmed example: the coal machine's **"lid"** is *not* in `Machine.NounsForMatching` (`["machine","dryer"]`) yet `open lid` works, because it's matched inline (`action.Match(["open"], ["lid"])`). So each entry below must be checked against both `NounsForMatching` **and** any inline `Match`/location handlers before being added. Treat the lists as candidates, not a to-do.

Also note: many missing words are purely **cosmetic alternates** a player rarely types (casket/cup/blade/hemp). The high-value subset is called out first.

---

## ZORK — gameplay-relevant gaps (verify, then likely add)

| Object | C# has | Missing (ZIL `SYNONYM`) | Note |
|---|---|---|---|
| Dam / Gate | `gate` | **`dam`, `gates`** | Verified real — `Gate.cs` only has `"gate"`; "examine dam"/"open gates" fail. The original `DAM` object is `SYNONYM DAM GATE GATES`. |
| Leaflet (`ADVERTISEMENT`) | `leaflet` | **`mail`, `booklet`, `advertisement`** | Early-game item in the mailbox; "read mail" is natural. |
| Map | `map` | **`parchment`** | It's described as "ancient parchment which appears to be a map." |
| Skeleton (`BONES`) | `skeleton` | `bones`, `body` | "examine bones". (Relates to #36.) |

## ZORK — cosmetic alternates (lower priority)

`WATER` →liquid,h2o · `GHOSTS` →fiends,force · `SKULL` →head · `LUNCH` →food,dinner · `BLACK BOOK` →prayer,page · `CHALICE` →cup · `TRIDENT` →fork · `CYCLOPS` →eye · `TRAP-DOOR` →cover · `BOTTLE` →container · `COFFIN` →casket · `PUMP` →air-pump · `KNIFE` →knives,blade · `MACHINE` →pdp10,lid *(lid is a FALSE POSITIVE — already works)* · `MAILBOX` →box · `MIRROR` →reflection · `PAINTING` →art,canvas · `LEAVES` →leaf · `INFLATABLE-BOAT` →valve · `ROPE` →hemp,coil · `SCREWDRIVER` →driver · `COAL` →heap · `SCARAB` →bug,beetle · `SWORD` →blade · `THIEF` →person · `TUBE` →tooth,paste · `ENGRAVINGS` →inscription · `SAILOR` →aviator

## PLANETFALL — gameplay-relevant gaps (verify, then likely add)

| Object | C# has | Missing (ZIL `SYNONYM`) | Note |
|---|---|---|---|
| Chemical fluid | `liquid`, `quantity` | **`chemical`, `fluid`** | The comm-room coolant puzzle; "pour chemical"/"examine fluid". |
| Mutants (rat-ant, troll, grue, triffid) | `mutant`, names | **`monster`** | In the Bio Lab chase the player will type "attack monster" / "kill monster". |
| Lazarus breastplate (`LAZARUS-PART`) | `plate` | **`breastplate`, `lazarus`, `part`** | Floyd's emotional item; "give breastplate" / "show lazarus" fail. |
| Access cards (all) | `card` | **`cards`** (plural) | "examine cards" / "take cards" with several in hand. |
| Lamp | `lamp`, `light` | `lantern` | Zork-muscle-memory players type "lantern". |
| Print-out | `paper`, `output` | `printout` | "read printout". |

## PLANETFALL — cosmetic alternates (lower priority)

`PADLOCK` →pad · `HIGH-PROTEIN` →fluid,food · `CARTON` →carton · boards →`boards` (plural) · `PLIERS` →pair · `FLOYD`/`DEAD-FLOYD` →companion · `COMM-SCREEN` →communications,message · `CONTROLS` →dials,gauges · `SCRUB-BRUSH` →scrubber · `PATROL/LAB-UNIFORM` →suit · `CELERY` →stalk · `BROCHURE` →pamphlet,leaflet · `SAFETY-WEB` →net,mass · `FOOD-KIT` →provisions · goo →blob,food (+pie/stew/beans) · `MEDICINE-BOTTLE` →label · `ACHILLES` →achilles · `ROBOT-HOLE` →hole,doorway · `ACCESS-PANEL`/`CUBE` →socket · spools →`spools` (plural) · `TERMINAL` →keyboard · `SPOOL-READER` →opening · `RELAY` →micro · `LASER` →device

## Adjectives (secondary axis)

The original also has `(ADJECTIVE …)` per object (e.g., the egg's SHARP/EGYPTIAN/ANCIENT/ENAMELED, the lamp's BRASS). The C# folds adjectives into multi-word `NounsForMatching` phrases ("brass lantern"), which covers the common cases but not free-standing adjective references. Lower priority than the synonyms above.

## Recommended approach

1. Start with the gameplay-relevant tables; for each, confirm the word isn't already handled (via `NounsForMatching` or inline `Match`) and add it to `NounsForMatching` if not.
2. Consider a small shared "treasure"/"monster"/"card(s)"-style synonym handling so plurals and category words ("monster", "cards") resolve generally.
3. Leave the cosmetic tail unless doing a completeness pass.
4. Add a couple of regression tests (e.g., `examine dam`, `attack monster`, `give breastplate`) in the same style as the #23 sceptre tests.

_Found via a mechanical ZIL `SYNONYM`/`ADJECTIVE` vs `NounsForMatching` diff (the systematic generalization of #23)._

arsindelve:
Fixes #228.

## What & why

The parser's noun vocabulary is **reflected from every item's `NounsForMatching`** (`Repository.GetNouns`). So any `(SYNONYM …)` word the original Infocom source accepted but the port had dropped silently fails to resolve — the parser never even produces an intent for it. This is the systematic generalization of the already-fixed #23 (sceptre vs. scepter).

Per the issue's caveat that the mechanical diff **over-reports**, each candidate was checked against both `NounsForMatching` **and** any inline `Match`/location handlers before being added. The high-value, gameplay-relevant gaps were restored; the cosmetic tail was left for a future completeness pass.

## Changes

**Zork I**
| Object | Added |
|---|---|
| `Leaflet` (ZIL `ADVERTISEMENT`) | `mail`, `booklet`, `pamphlet`, `advertisement` — "read mail" from the mailbox now works |
| `Map` | `parchment` / `ancient parchment` — it's "ancient parchment which appears to be a map" |
| `Skeleton` (ZIL `BONES`) | `bones`, `body` |
| Dam | new examinable **`FloodControlDam`** item (see below) |

**Planetfall**
| Object | Added |
|---|---|
| Mutants — grue, troll, rat-ant, triffid | `monster` (shared via new `MutantBase`) |
| `Lamp` | `lantern` (Zork muscle memory) |
| `ComputerOutput` | `printout` — described throughout as "the printout" |

> **Scope note on `monster`:** this restores *vocabulary*, not combat. The four mutants don't implement `ICanBeAttacked`, so adding `monster` makes the noun **resolve** (and disambiguate among the four, or examine) instead of failing as an unknown word — `attack monster` still gives the same generic "you can't attack that" response the game already gave for `attack troll`. The shared `monster` noun lives on a new `MutantBase` (de-duplicating `Size`/`ExaminationDescription` too) and is excluded from `NounsForPreciseMatching`, mirroring the existing `FromitzBoardBase`/`AccessCard` pattern for shared category nouns.

### The dam needed special handling
The issue's first row conflates two different gates: the C# `Gate` **item** is the gate at **Entrance to Hades**, not the dam's sluice gates. In the original, `DAM` (`SYNONYM DAM GATE GATES`) is an examinable object, but this port models the dam as a **location** with no object carrying those nouns — so "examine dam"/"examine gates" couldn't even parse. Rather than polluting the unrelated Hades gate, I added a fixed, non-takeable `FloodControlDam` item placed at the `Dam` location whose examination text reuses `Dam.GatesDescription()` to report the live sluice-gate state.

### False positives confirmed and deliberately skipped
- The medical robot breastplate **already** has `breastplate` (issue claimed only `plate`).
- The comm-room coolant **already** matches `chemical`/`fluid` inline in `CommRoom`.
- The coal machine's `lid` **already** works inline (called out in the issue).

## Tests
Each restored synonym has a test living in the **existing topical class** for that item (no dedicated synonym-test class): Zork's map/leaflet/skeleton go in `MiscItemTests`, the dam in `DamTests`; Planetfall's mutants in `BioLabEscapeTests`, the lamp in `LabsAndDoorsTests`, the computer output in `ComputerRoomTests`. They exercise the real `GetResponse` pipeline (e.g. `read mail` → leaflet text, `examine dam` → "sluice gates") rather than only checking the noun list. Full suites green:
- `ZorkOne.Tests`: 1221 passed
- `Planetfall.Tests`: 2228 passed
- `UnitTests`: 825 passed

---
_Generated by [Claude Code](https://claude.ai/code/session_01QpPS3YuHFBG2VGgARAtNuC)_


Planetfall/Item/Lawanda/BioLab/Triffid.cs
```diff
@@ -2,7 +2,9 @@ namespace Planetfall.Item.Lawanda.BioLab;
 
 public class Triffid : ItemBase, ICanBeExamined
 {
-    public override string[] NounsForMatching => ["triffid", "plant"];
```
arsindelve, line 5: Create a base class for these, and carefully examine and implement how we share nouns on the base class by looking at the fromitz board or the Beddit or

Planetfall.Tests/SynonymVocabularyTests.cs
arsindelve, line 1: Did you need to make another class, or was there an existing test class where this would have belonged?

Planetfall.Tests/SynonymVocabularyTests.cs
```diff
@@ -0,0 +1,65 @@
...
+
+namespace Planetfall.Tests;
+
+/// <summary>
+/// Issue #228: the C# port's NounsForMatching arrays had dropped some of the original
+/// Infocom (SYNONYM ...) vocabulary, so perfectly reasonable references silently failed to
+/// resolve. These tests pin the restored gameplay-relevant synonyms for Planetfall.
+/// </summary>
+public class SynonymVocabularyTests : EngineTestsBase
+{
```
arsindelve: Why not use test cases here like in Zork?

Issue #230, opened 2026-06-21
GET session-state endpoint returns an empty inventory while items are held (web reconnect shows empty hands)


arsindelve:
## Summary

The `GET /ZorkOne?sessionId=…` endpoint returns a `GameResponse` whose `inventory` field is **empty (`[]`)** even when the player is holding items. The equivalent `POST /ZorkOne` at the identical game state returns the correct inventory. Because the web client uses **GET to rehydrate on reconnect/refresh**, a returning player sees an empty inventory while their items are in fact still held. The underlying saved state is intact — only the structured `inventory` projection in the GET response is wrong.

**Severity:** Medium · **Deterministic** (no RNG/AI involved) · reproduced live against prod.

## Steps to reproduce (live, prod Zork)

From a session that has progressed into the Living Room holding the brass lantern (e.g. open mailbox → S → E → open window → in → W → take lamp):

- **POST** `/ZorkOne` body `{ "input": "inventory", "sessionId": "<id>", "noGeneratedResponses": true }`
  → envelope: `inventory: ["lantern"]`  ✅
- **GET** `/ZorkOne?sessionId=<id>`
  → envelope: `inventory: []`  ❌ (prose/state still reflect the lantern is held)

Same session, same moment, two different answers.

## Root cause

The engine maintains a flat `Inventory` projection (`List<string>`) that is **separate** from the authoritative game state in `Context.Items`:

- `GameEngine/GameEngine.cs:57` — `public List<string> Inventory { get; set; }`
- It is initialized empty at `GameEngine.cs:98` / `:121`.
- It is only **recomputed from `Context.Items`** in:
  - `GameEngine.cs:581` — after a turn is processed in `GetResponse(...)`
  - `GameEngine.cs:412` — in `RestartAfterDeath(...)`

`RestoreGame(...)` (`GameEngine/GameEngine.cs:364-380`) rebuilds `Context` from the saved blob but **never refreshes the `Inventory` projection**:

```csharp
public IContext RestoreGame(string data)
{
    var deserializeObject = JsonConvert.DeserializeObject<SavedGame<TContext>>(data, JsonSettings());
    ...
    Repository.Restore(allItems, allLocations);
    Context = deserializeObject.Context ?? throw new ArgumentException();
    Context.Engine = this;
    Context.Game = new TInfocomGame();
    return Context;          // <-- Inventory projection left untouched (stays [])
}
```

[...]

PR #234, opened 2026-06-21
Stop the narrator inventing objects on deflection responses

arsindelve:
## Problem

The AI narrator fabricates concrete objects that don't exist in the room. Found live by the AdventureBreaker harness: deflecting `examine the wizard standing in the corner` in the **Studio**, the narrator replied *"…it's just a paint-splattered broom."* There is **no broom anywhere in Zork I** (verified: `grep -rin broom ZorkOne/` → no matches; the Studio only `StartWithItem<Manual>` at `ZorkOne/Location/Studio.cs:58`). A player could then `take broom` and fail — a trust-breaking hallucination.

## Root cause

The three "negative" deflection prompts —
`NounNotPresentOperationRequest`, `VerbHasNoEffectOperationRequest`, `CommandHasNoEffectOperationRequest` —
gave the model **no instruction to stay inside the room state** and ran at the base **`Temperature = 0.8`**, which is tuned for creative narration and rewards embellishment on exactly these responses. (The model already receives the real room contents via `GetDescriptionForGeneration()`; it just wasn't constrained to them.)

The codebase already has the idiom for this: `CannotGoThatWayRequest` and the sub-location prompts say *"Do not be creative about why… do not alter the state of the game or provide additional information"* (see `RequestTests.cs`). These deflection prompts simply never got the same guard.

## Fix

- Add an explicit guard to all three prompts: do **not** invent/name/substitute any object, item, scenery, exit, or character that isn't already present in the supplied location description; refer only to what's actually there.
- Lower `Temperature` to `0.4` on these deflections so they don't embellish.

This is per-response (it doesn't dampen normal creative narration) and mirrors the existing convention. A complementary global rule could live in the system prompt (AWS Secrets Manager), but that's config, not code, so it's out of scope here.

## Tests (TDD)

The end-to-end symptom is **stochastic LLM output**, which can't be a deterministic unit test. The deterministic seam is the **prompt contract**, tested in the existing `UnitTests/Models/RequestTests.cs`: three new cases assert the guard text is present and `Temperature < 0.8`. They are **red before** this change (confirmed: failing on the missing `"Do not invent"` clause) and **green after**.

No regressions — each suite run separately:
- `UnitTests` — 828 passed, 1 skipped
- `ZorkOne.Tests` — 1212 passed
- `Planetfall.Tests` — 2219 passed

Closes the `narrator_hallucination` finding from the AdventureBreaker run. Related inventory bug tracked in #230.

https://claude.ai/code/session_01DsEMy2gYe2qexc95aJR4KR

---
_Generated by [Claude Code](https://claude.ai/code/session_01DsEMy2gYe2qexc95aJR4KR)_

### DIFF
--- a/Model/AIGeneration/Requests/CannotGoThatWayRequest.cs
+++ b/Model/AIGeneration/Requests/CannotGoThatWayRequest.cs
@@ -8,5 +8,8 @@ public class CannotGoThatWayRequest : Request
             $"The player is in this location: \"{location}\". They tried to go {direction} but going that way is  " +
             $"not possible from this location. Respond with a very short, sarcastic and simple message informing them " +
             $"that they cannot go that way. Do not be creative about why or what is preventing them, do not alter the state of the game or provide additional information. ";
+
+        // Same "stay in room state" deflection category as the no-effect prompts: keep creativity low.
+        Temperature = DeflectionTemperature;
     }
 }
--- a/Model/AIGeneration/Requests/CommandHasNoEffectOperationRequest.cs
+++ b/Model/AIGeneration/Requests/CommandHasNoEffectOperationRequest.cs
@@ -21,11 +21,10 @@ public class CommandHasNoEffectOperationRequest : Request
             Player input: ""break the window""
             AI response: ""You consider breaking the window, but then think better of it. Maybe there's another way to solve this problem.""
 
-            Please follow these guidelines when generating responses. Do not invent, name, or describe any object, item,
-            scenery, exit, or character that is not already explicitly present in the location description above; refer only to
-            things actually present, or keep it generic, and do not alter the state of the game or add new information. ";
+            Please follow these guidelines when generating responses. " + NoInventionGuard + @" When you describe a
+            benign or atmospheric action creatively (like dancing or singing), that creativity applies to tone and flavor only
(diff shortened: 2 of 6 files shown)


arsindelve:
Thanks for the review — addressed all four points in `9df41be`:

1. **`CommandHasNoEffect` contradiction** — added an explicit clarification that for the benign/atmospheric branch ("describe creatively"), the creativity is **tone/flavor only**, never introducing concrete objects, characters, exits, or state. So "creative" and the anti-invention guard no longer pull against each other.
2. **`CannotGoThatWayRequest` inconsistency** — it now uses the same low deflection temperature, so all the "stay in room state" prompts behave consistently rather than leaving that one at the 0.8 default.
3. **Magic-number / guard duplication** — extracted `Request.NoInventionGuard` and `Request.DeflectionTemperature` on the base class; the four deflection prompts now reference the shared constants so they move together when tuned.
4. **Trailing newlines** — added to the touched request files.

New/updated tests: a `CannotGoThatWayRequest` low-temperature case (red before / green after) and a `"tone and flavor only"` assertion on `CommandHasNoEffect`. UnitTests (829), ZorkOne.Tests (1212), and Planetfall.Tests (2219) all pass locally.

---

PR #242, opened 2026-06-22
Fix #241: god mode no longer rebuilds the repository without Init()

### LINKED ISSUE #241: God mode rebuilds the repository without Init() — returns empty containers and discards live game state
## Summary

`Repository.LoadAllItems` and `Repository.LoadAllLocations` — both called by the god-mode commands (`god mode take` / `god mode go`) — **replace the entire repository dictionary with fresh `Activator.CreateInstance` instances and never call `Init()`**. As a result:

1. Items come back **without their `Init()`-established state**. `god mode take cardboard box` on a fresh prod run yields an **empty box**, even though `CardboardBox.Init()` does `StartWithItemInside<GoodBedistor>()` + the two megafuses + `CrackedFromitzBoard`.
2. Replacing the whole dictionary **discards the restored state of every other item/location**, scrambling an in-progress session (containers empty out, flags/positions reset to defaults).

This makes god mode unsafe for testing — any item or container whose initial contents/state come from `Init()` can't be reached or verified through god mode.

## Repro (live prod, fresh session)

```
god mode take cardboard box
examine the cardboard box        -> "The cardboard box is empty."   (should contain a good bedistor, two megafuses, a cracked fromitz board)
```

## Root cause (`GameEngine/Repository.cs`, verified in `origin/main` @ `c31e9ec`)

```csharp
public static void LoadAllItems(string gameName = "ZorkOne")
{
    lock (_allItems)
    {
        _allItems = new Dictionary<Type, IItem>();           // <-- throws away the current (restored) items
        var types = Assembly.Load(gameName).GetTypes();
        foreach (var type in types)
            if (type is { IsClass: true, IsGenericType: false, IsAbstract: false } &&
                type.IsSubclassOf(typeof(ItemBase)))
            {
                var instance = (ItemBase)Activator.CreateInstance(type)!;   // <-- no Init() call
                _allItems.Add(type, instance);
            }
    }
}
```

`LoadAllLocations` has the same shape. Contrast `GetLocation<T>()`, which **does** call `location.Init()` when it lazily creates a location — so the normal play path is fine; only the bulk-rebuild path skips `Init()` and clobbers existing instances.

Because a deployed Lambda turn does `RestoreGame` (→ `Repository.Restore(...)`) at the start, a god-mode command then overwrites that restored state with default, un-`Init()`-ed instances before the turn is serialized back — persisting the scrambled state.

## Suggested fix

- Call `Init()` on each instance created in `LoadAllItems` / `LoadAllLocations`, and
- prefer **adding only missing types** to the existing dictionaries rather than replacing them wholesale, so already-loaded/restored instances (and their live state) are preserved.

## Notes

Found via the AdventureBreaker harness while validating a suspected "bedistor noun collision." That collision turned out to be a **god-mode artifact too**: god mode resolves nouns through `Repository.GetItem` (raw containment, first match), which lacks the `NounsForPreciseMatching` + `MatchNounAndAdjective` disambiguation the real `SimpleInteractionEngine` parser uses — so `god mode take good bedistor` grabbed the *fused* bedistor. The real player parser disambiguates correctly (it would prompt *"Do you mean the good ninety-ohm bedistor or the fused ninety-ohm bedistor?"*), so there is **no player-facing bedistor bug** — only these god-mode issues.

arsindelve:
## Summary

Fixes #241. `Repository.LoadAllItems` / `LoadAllLocations` — the bulk loaders behind the god-mode commands (`god mode take` / `god mode go`) — replaced the entire repository dictionary with fresh `Activator.CreateInstance` instances and **never called `Init()`**. This broke god mode two ways:

1. **Containers came back empty.** `god mode take cardboard box` yielded an empty box because `CardboardBox.Init()` (which stocks it with a good bedistor, two megafuses, and a cracked fromitz board) never ran.
2. **Live/restored state was clobbered.** Replacing the whole dictionary discarded the restored state of every other item and location, scrambling an in-progress session — and, because a deployed Lambda turn does `RestoreGame` at the start, persisting that scrambled state.

## Fix

Both loaders now follow the issue's suggested fix and mirror the existing single-instance lazy-load paths (`GetItem<T>()` / `GetLocation<T>()`):

- **Add only missing types** instead of replacing the dictionary, so already-loaded/restored instances and their live state are preserved.
- **Call `Init()` on newly created instances** — containers via `ICanContainItems.Init()`, locations via `Init()`.

The now-redundant `location.Init()` in `GodModeProcessor.Go()` is left in place; it's harmless because `LocationBase.ItemPlacedHere` dedupes by type.

Per the issue's notes, the bedistor "noun collision" is **not** touched — it's a god-mode-only artifact (god mode resolves nouns through raw `Repository.GetItem`, which lacks the `NounsForPreciseMatching` + `MatchNounAndAdjective` disambiguation the real `SimpleInteractionEngine` parser uses), not a player-facing bug.

## Tests (TDD)

Added three tests to `UnitTests/RepositoryTests.cs`, confirmed **red** before the fix and **green** after:

- `LoadAllItems_InitializesContainers_SoCardboardBoxStartsWithItsContents`
- `LoadAllItems_PreservesExistingInstancesAndTheirState`
- `LoadAllLocations_PreservesExistingInstances`

No regressions:

- `UnitTests`: 833 passed, 0 failed
- `ZorkOne.Tests`: 1226 passed, 0 failed (exercises the existing `LoadAllItems` consumer in `MiscItemTests`)
- `Planetfall.Tests`: 2231 passed, 0 failed


GameEngine/Repository.cs
```diff
@@ -278,15 +284,24 @@ public static void LoadAllItems(string gameName = "ZorkOne")
...
 
                 if (type is { IsClass: true, IsGenericType: false, IsAbstract: false } &&
                     type.IsSubclassOf(typeof(ItemBase)))
                 {
+                    // Issue #241: only add items we're missing - never replace the whole
+                    // dictionary, or we would discard the live/restored state of an in-progress
+                    // session. Containers get Init()'d so they start with their contents,
+                    // mirroring GetItem<T>(). (A container's Init() may lazily create the items
+                    // it holds via GetItem<T>(); those then get skipped here as already-present.)
+                    if (_allItems.ContainsKey(type))
```
arsindelve, line 305: `ContainsKey → continue` is exactly right for preserving restored state. One subtle consequence of no longer resetting the dictionary: `LoadAllItems`/`LoadAllLocations` now **accumulate across games** — calling this for `ZorkOne` and then `Planetfall` in the same process without a `Repository.Reset()` in between would leave both games' items loaded. Harmless in practice (the Lambdas are per-game and the tests `Reset()` first), but a one-line comment noting the single-game assumption would save a future head-scratch.

---

UnitTests/RepositoryTests.cs
arsindelve, line 1: Minor: this file still ends without a trailing newline (`\ No newline at end of file` in the diff). The repo has been normalizing these lately (e.g. the trailing-newline pass on `Verbs.cs`/`DomeTests.cs`) — worth adding one here too.

---

arsindelve (commented):
## Looks correct — no blocking issues ✅

Reviewed against the surrounding code (not just the diff). The fix is minimal and, importantly, **mirrors the canonical lazy-load paths exactly**:

- `GetItem<T>()` itself only `Init()`s containers (`if (item is ICanContainItems container) container.Init()`), so `LoadAllItems` doing the same is **consistent, not a gap** — non-container items never get `Init()` on the normal path either.
- `GetLocation<T>()` `Init()`s locations → `LoadAllLocations` now matches.

What I verified holds up:
- **Empty-container bug fixed** (containers `Init()` → `CardboardBox` gets its contents) and **state preserved** (`ContainsKey → continue` instead of replacing the dict).
- **Iteration safety:** the loop walks `Assembly.GetTypes()` (a fixed array), not `_allItems`, while a container's `Init()` can add its contents to `_allItems` mid-loop; the `ContainsKey` guard then skips them, avoiding a duplicate-key `Add`. 👍 for calling this out in the comment.
- **Lock re-entrancy:** `Init()` → `GetItem<T>()` mutates `_allItems` while `lock(_allItems)` is held, but it's same-thread/single-threaded with no concurrent dict iteration. Safe.
- **Blast radius:** the only *production* caller of these bulk loaders is `GodModeProcessor` (a debug command); everything else is the lazy-load path or tests.
- **Tests** are genuine red→green (traced each): pre-fix, replace-the-dict breaks `BeSameAs` and the un-`Init`'d box fails `NotBeEmpty`; `InitializesContainers` also implicitly exercises the mid-loop `Init()`-adds-items path.

Two minor inline notes below (cross-game accumulation comment + trailing newline), neither blocking.

One more non-blocking note (file not in the diff): the `location.Init()` left in `GodModeProcessor.Go()` is now redundant for the not-yet-loaded case (the target gets `Init()`'d once by `LoadAllLocations` and again by `Go()`). It's harmless — `LocationBase.ItemPlacedHere` dedupes by type ("No duplicates in any game, ever.") and the second `Init()` is a no-op on already-placed items — but it could be dropped for clarity.

Good fix. Ship it.

---

GameEngine/Repository.cs
arsindelve: Good call — added a comment to both LoadAllItems and LoadAllLocations noting they accumulate across games and assume one game per process (the Lambdas are per-game), so two games without a Repository.Reset() in between would leave both loaded.

UnitTests/RepositoryTests.cs
arsindelve: Added the trailing newline to UnitTests/RepositoryTests.cs.

Issue #244, opened 2026-06-22
Take/drop/put noun resolution ignores adjectives — "take the good bedistor" acts on the fused bedistor (examine resolves it correctly)


arsindelve:
## Summary

`Repository.GetItemInScope` (the resolver behind take / drop / put) ignores adjectives: when the **good** and **fused** bedistor are both in scope, the reference **"good bedistor"** resolves to the **fused** one. Cause: `ItemBase.HasMatchingNoun` falls back to word-boundary **containment**, so the fused bedistor's bare noun `"bedistor"` matches the input `"good bedistor"`; and `GetItemInScope` searches the **location first**, returning that match — with no adjective precision. The **examine** path (`SimpleInteractionEngine` → `MatchNounAndAdjective`) honors the adjective, so the two disagree.

**Severity:** Low–Medium. **Deterministic.**

## Corrected impact (this is NOT a puzzle blocker)

The course-correction puzzle **is solvable** — the existing `CourseControlTests.FullSolution` passes using **`put good in cube`** (and `take fused with pliers`). So the intended solution works.

The bug bites the **fuller, equally natural phrasings**, which the prod AI parser readily produces from input like *"take the good bedistor"* / *"put the good bedistor in the cube"*:

- `take the good bedistor` (prod) → *"It seems to be fused to its socket."* (acted on the fused)
- `put the good bedistor in the cube` (prod) → *"You don't have the fused ninety-ohm bedistor."*

So a player who says "the good bedistor" by name gets the wrong item, even though "good" alone works. It also generalizes to any two items sharing a base noun but differing by adjective.

## Confirmed root cause (deterministic)

With the good bedistor in inventory and the fused in the open cube (`main` @ `341a64b`):

| call | returns |
|---|---|
| `GetItemInScope("good bedistor", ctx)` | **FusedBedistor** ✗ |
| `GetItemInScope("good ninety-ohm bedistor", ctx)` | **FusedBedistor** ✗ |
| `GetItemInScope("good", ctx)` | `null` (no in-scope item has the bare noun "good") |

`GetItemInScope` searches the location before inventory and returns the first `HasMatchingNoun` hit, and `HasMatchingNoun`'s containment fallback (`" good bedistor ".Contains(" bedistor ")`) makes the fused match. Examine's adjective-aware path doesn't.

## TDD — test the resolver, not `GetResponse`

[...]

PR #249, opened 2026-06-22
Fix #246: multi-noun (put/give/slide X in/to Y) noun resolution now honors adjectives

### LINKED ISSUE #246: Multi-noun (put/give/slide X in/to Y) still ignores adjectives — resolves "kitchen card" to the wrong card (#244 only fixed single-noun)
## Summary

The #244 fix made **single-noun** take/drop/put (`Repository.GetItemInScope`) adjective-aware — confirmed working across every collision family (bedistor, batteries, megafuses, cards, spools, fromitz, goo). But the **multi-noun** path (`put X in Y`, `give X to Y`, `slide X through Y`) resolves its nouns through a **different** method that the fix didn't touch, so it is **still adjective-blind**: an adjective-qualified noun like "kitchen card" resolves to the wrong item when more than one candidate is in scope.

This is the **cards-into-slots** puzzle pattern (`slide the kitchen card through the slot`, `put the shuttle card in the slot`).

**Severity:** Medium. **Deterministic** (order-dependent). Confirmed against `main` @ `8175684` (post-#245).

## Confirmed repro (deterministic, resolver seam)

Two cards carried, shuttle first in inventory:

```csharp
Take<ShuttleAccessCard>();
Take<KitchenAccessCard>();
context.HasMatchingNoun("kitchen card");   // -> ShuttleAccessCard  (BUG; should be Kitchen)
```

`context.HasMatchingNoun` falls back to word-boundary containment (`"kitchen card"` contains `"card"`, a Shuttle noun) and returns the **first** inventory match — ignoring the "kitchen" adjective. (Note this path is **inventory-first**, where `GetItemInScope` was location-first, so the failing scope config differs, but the root cause is the same.)

## Root cause (`GameEngine/IntentEngine/MultiNounEngine.cs`, `main` @ `8175684`)

`IsItemHere` resolves each noun via the raw matcher, not the fixed `GetItemInScope`:

```csharp
private static (bool IsHere, IItem? item) IsItemHere(IContext context, string item)
{
    var result = context.HasMatchingNoun(item);                 // <-- raw containment, no adjective
    if (result.HasItem) return result;
    result = context.CurrentLocation.HasMatchingNoun(item);     // <-- same
    ...
}
```

There's an extra twist: `MultiNounEngine`'s own disambiguation (built from `NounsForPreciseMatching`) is adjective-aware, so a *specific* "kitchen card" passes the ambiguity check with a single match — and then `IsItemHere` mis-resolves it anyway. So, like the original #244 (examine vs take), the **disambiguation check and the actual resolution disagree**, now *within* the multi-noun path.

## Suggested fix

Route `MultiNounEngine.IsItemHere` through the same adjective-aware resolver as the single-noun fix (`GetItemInScope` / `GetPreciseMatchInScope`) so precise/adjective matches beat the containment fallback. The robust version is to make **all** scope resolution share one adjective-aware helper — `context.HasMatchingNoun` / `location.HasMatchingNoun` are raw and likely have other adjective-blind callers; #244 patched only `GetItemInScope`.

**TDD (deterministic, no AI):** the seam test above is red today. Add it next to the bedistor multi-noun coverage; assert `context.HasMatchingNoun("kitchen card")` (or, once routed, `GetItemInScope`) returns the `KitchenAccessCard` regardless of inventory order, and that a multi-noun `put`/`slide` with two cards acts on the named one.

## Notes

Found by the AdventureBreaker harness while validating the deployed #244 fix across the card/uniform/goo families. The single-noun families are all fixed; this is the residual multi-noun path. (End-to-end the trigger is the prod AI parser producing "kitchen card"; god mode can't set up two distinct cards because `Repository.GetItem` — used by god mode — is also raw-matched, a separate debug-only quirk.)

arsindelve:
## Summary

Closes #246.

The #244 fix made **single-noun** take/drop/put adjective-aware (`Repository.GetItemInScope`), but the **multi-noun** path (`put X in Y`, `give X to Y`, `slide X through Y`) resolves its nouns through a different method — `MultiNounEngine.IsItemHere` — which the fix never touched. `IsItemHere` used the raw, adjective-blind containment matcher (`context.HasMatchingNoun` / `location.HasMatchingNoun`), so an adjective-qualified noun like `"kitchen card"` or `"good bedistor"` still resolved to the wrong item when more than one candidate was in scope.

Root cause is the same disagreement #244 fixed, now *within* the multi-noun path: `ItemBase.HasMatchingNoun` falls back to word-boundary containment, so a shuttle card's bare noun `"card"` is contained in `"kitchen card"` (and a fused bedistor's `"bedistor"` in `"good bedistor"`), and the **first** such match in inventory order won — ignoring the adjective. `MultiNounEngine`'s own disambiguation is precise (built from `NounsForPreciseMatching`), so a specific noun passes the ambiguity check with a single match and then `IsItemHere` mis-resolves it anyway.

## Fix

- Promote `Repository.GetPreciseMatchInScope` from `private` to a **shared public** adjective-aware resolver (with a null/normalization guard), and document it as the one helper every scope resolver runs first.
- Run that precise pass at the **start** of `MultiNounEngine.IsItemHere`, before the existing inventory-first containment fallback. An exact match against an item's `NounsForPreciseMatching` now beats a containment match in the multi-noun path too.
- The existing inventory-first containment + description fallback is left **untouched**, so plain non-adjective nouns (`"take bedistor"`, `"put leaflet in mailbox"`, etc.) behave exactly as before.

## Tests (TDD)

- **`CourseControlTests.PutGoodBedistorInCube_WhenFusedIsFirstInInventory_ActsOnTheNamedGoodBedistor`** — proving test, **red before / green after**. With the fused bedistor first in inventory, `put good in cube` used to put the *fused* bedistor in the cube (response `"Done."`, Course Control not fixed); it now acts on the named *good* bedistor (`"The warning lights go out…"`, Course Control fixed).
- **`CardTests.KitchenCard_ResolvesToKitchenNotShuttle_RegardlessOfInventoryOrder`** — guards the issue's named scenario: `"kitchen card"` resolves to the kitchen card via the shared resolver regardless of inventory order.

## Verification

- `dotnet test UnitTests` → 834 passed, 1 skipped
- `dotnet test ZorkOne.Tests` → 1226 passed
- `dotnet test Planetfall.Tests` → 2235 passed (includes the 2 new tests)
- `dotnet build Zork.sln` → 0 warnings, 0 errors

https://claude.ai/code/session_01Q1gWczxJHGfKHSwKG4Vjis

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q1gWczxJHGfKHSwKG4Vjis)_

### DIFF
--- a/GameEngine/IntentEngine/MultiNounEngine.cs
+++ b/GameEngine/IntentEngine/MultiNounEngine.cs
@@ -139,6 +139,13 @@ public class MultiNounEngine : IIntentEngine
         // giving the shared adjective-aware pass priority over the containment fallback below. The
         // existing inventory-first fallback is left untouched so plain, non-adjective nouns behave
         // exactly as before.
+        //
+        // Note the precise pass searches room-before-inventory (matching GetItemInScope), while the
+        // fallback below stays inventory-before-room - intentional, and not a contradiction to
+        // "fix": the two orders can only disagree when one noun is an exact precise-noun match on
+        // two different in-scope items at once, and Process() runs CheckDisambiguation (over the
+        // same scope) before we ever get here, so that collision is already turned into a "do you
+        // mean..." prompt rather than a silent pick.
         var preciseMatch = Repository.GetPreciseMatchInScope(item, context);
         if (preciseMatch is not null)
             return (true, preciseMatch);
--- a/GameEngine/IntentEngine/MultiNounEngine.cs
+++ b/GameEngine/IntentEngine/MultiNounEngine.cs
@@ -133,6 +133,16 @@ public class MultiNounEngine : IIntentEngine
 
     private static (bool IsHere, IItem? item) IsItemHere(IContext context, string item)
     {
+        // Issue #246: an adjective-qualified noun ("kitchen card", "good bedistor") must resolve to
+        // the precise item, not the first raw-containment match (a shuttle "card" is contained in
+        // "kitchen card"; a fused "bedistor" in "good bedistor"). Mirror the single-noun #244 fix by
+        // giving the shared adjective-aware pass priority over the containment fallback below. The
+        // existing inventory-first fallback is left untouched so plain, non-adjective nouns behave
+        // exactly as before.
+        var preciseMatch = Repository.GetPreciseMatchInScope(item, context);
+        if (preciseMatch is not null)
+            return (true, preciseMatch);
+
         var result = context.HasMatchingNoun(item);
         if (result.HasItem)
             return result;
--- a/GameEngine/Repository.cs
+++ b/GameEngine/Repository.cs
@@ -125,10 +125,21 @@ public static class Repository
     /// <see cref="IItem.NounsForPreciseMatching"/> contains an exact match for the noun. This is
     /// the adjective-aware pass that lets "good bedistor" win over a bare "bedistor" containment
     /// match (issue #244). Returns null when no precise match is in scope, leaving the broader
-    /// containment fallback in <see cref="GetItemInScope"/> to do its job.
+    /// containment fallback in the caller to do its job.
+    ///
+    /// Shared by every scope resolver so they agree on adjectives: <see cref="GetItemInScope"/>
+    /// (single-noun take/drop/put, #244) and <see cref="IntentEngine.MultiNounEngine"/>'s noun
+    /// resolution (multi-noun put/give/slide, #246). The raw <see cref="IItem.HasMatchingNoun"/> /
+    /// location matchers are adjective-blind containment matchers, so callers must run this pass
+    /// first to keep "kitchen card" from resolving to a shuttle "card".
     /// </summary>
-    private static IItem? GetPreciseMatchInScope(string noun, IContext context)
+    public static IItem? GetPreciseMatchInScope(string? noun, IContext context)
     {
+        if (string.IsNullOrEmpty(noun))
+            return null;
+
+        noun = noun.ToLowerInvariant().Trim();
+
         // Room before inventory, mirroring the search order in GetItemInScope. Guard against null:
         // production always returns a real list, but under-configured test mocks of IContext /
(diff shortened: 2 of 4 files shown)


arsindelve:
Thanks for the thorough review. Acting on the test-coverage point in `9f107bc`:

You're right that `CardTests.KitchenCard_ResolvesToKitchenNotShuttle...` asserts the resolver helpers directly rather than driving the engine — it's green before and after the fix, so as written its comment overclaimed. I've reworded it to state accurately that it's a **resolver / `NounsForPreciseMatching`-definition guard**, and that the red-before/green-after **engine-path** proof is `CourseControlTests.PutGoodBedistorInCube_WhenFusedIsFirstInInventory...` (which goes through `GetResponse("put good in cube")` → `IsItemHere` → `PutProcessor`).

On the suggested `"put kitchen card in slot"` / `"slide kitchen card through slot"` end-to-end test: I traced it and it **wouldn't exercise `IsItemHere` either**. `SlotBase.RespondToMultiNounInteraction` runs in the items-on-the-ground pass (`MultiNounEngine.Process`, before `ProcessNonLocationTwoItemInteraction`) and resolves the card **by type** — `action.Match<KitchenAccessCard, KitchenSlot>(...)` + `context.HasItem<KitchenAccessCard>()` — then returns a positive interaction and short-circuits. So the slide/put succeeds for the kitchen slot whenever a kitchen card is carried, regardless of inventory order or which other cards are present; `IsItemHere` is never reached on that path. The cube can't stand in for it either (`CanOnlyHoldTheseTypes => [typeof(BedistorBase)]`), which is why the engine-path proof lives in the bedistor scenario and the card collision is guarded at the resolver level.

Net: no production change — the three independent review passes I ran agreed the fix is correct (the order-flip concern is intercepted by `CheckDisambiguation`; the closed-container path is preserved by the untouched fallback). Just the test-comment accuracy fix.

---

PR #253, opened 2026-06-22
Fix #248: resolve "them" to item collections; keep pronoun antecedents across movement

### LINKED ISSUE #248: Pronoun antecedent too limited: \"them\" never resolves to a collection (take all → drop them), and movement clears \"it\" for carried items
## Summary

Pronoun resolution (`it` / `them`) is built on a single-noun-string antecedent (`context.LastNoun`) that can't represent a *set* of items and is **cleared** whenever the player moves or runs a multi-noun command. Two concrete, very natural failures result. Both confirmed deterministically against `main` @ `8175684` (Zork engine, no AI).

## A. "them" never resolves to a collection of items

After acting on multiple items, `them` fails — it only works for a single item that is intrinsically `IPluralNoun` (e.g. candles).

```
take all              ; drop them          -> "What item are you referring to?"
take all ; open case  ; put them in case   -> "What item are you referring to?"
examine all           ; take them          -> "What item are you referring to?"
take lantern ; take sword ; drop them       -> "What item are you referring to?"
```

`take all. put them in the case` is one of the most natural commands a player types (treasure hunting!), and it dead-ends.

**Root cause:** `context.LastNoun` holds one noun string; `ItProcessor.Check` only honors `them` when `Repository.GetItem(lastNoun) is IPluralNoun`. A collection of singular items has no single plural antecedent, so it falls through to the clarification prompt.

## B. Movement (and multi-noun) clears the antecedent — even for carried items

```
take lantern ; examine it ; east ; drop it
   -> LastNoun == ""   ->  "What item are you referring to?"
```

The lantern is still in the player's hand after walking east, but `it` no longer resolves.

**Root cause:** `MoveEngine.cs:36` and `MultiNounEngine.cs:29` both do `context.LastNoun = ""` unconditionally. Clearing on move makes sense for an object left behind, but not for an item the player is still carrying. (The multi-noun clear also means `take lamp. put it in the case. examine it` loses the antecedent after the `put`.)

## Suggested fix

Model the antecedent as **the set of items last referenced**, not a single noun string:
- `it` → the last single item; `them` → the last *set* (so `take all` / `take X and Y` populate it and `drop/put/take them` work).
- Resolve `them` against that set rather than requiring a single `IPluralNoun`.
- On movement, don't clear the antecedent for items that are still in scope (carried); only drop references to items left behind.

## TDD (deterministic, no AI)

Both repros are deterministic `GetResponse` chains — red today. Home: `UnitTests/GlobalCommands/ItProcessorTests.cs` (next to the existing `Them_SuccessPath` candles test). Suggested cases:
- `take all` then `drop them` → every taken item dropped (not the clarification prompt).
- `take lantern; take sword; drop them` → both dropped.
- `take lantern; examine it; <move>; drop it` → "Dropped" (antecedent survives a move while carried).

## Notes

Found by the AdventureBreaker harness. Existing tests cover the happy paths (single `it`, `them` for an `IPluralNoun`, most-recent tracking, no-antecedent clarification); these are the gaps. Separately, there is an AI pronoun resolver (`ResolvePronounsAsync`, uses `lastResponse`) that can bind `it` to a noun appearing only in narrator flavor text — flagging here as a related area to harden, but the two failures above are the deterministic, high-value ones.

arsindelve:
Fixes #248.

## Problem

Pronoun resolution was built on a single `context.LastNoun` string that can't represent a *set* of items, and movement cleared it unconditionally. Two natural commands broke (confirmed deterministically, no AI):

**A. `them` never resolved to a collection.** After acting on multiple items, `them` fell through to the clarification prompt:
```
take all              ; drop them   ->  "What item are you referring to?"
take lantern ; take sword ; drop them  ->  "What item are you referring to?"
```
It only worked for a single item that was intrinsically `IPluralNoun` (e.g. candles), because `ItProcessor.Check` resolved `them` against a single noun string.

**B. Movement cleared the antecedent even for carried items.**
```
take lantern ; examine it ; <go> ; drop it   ->  "What item are you referring to?"
```
`MoveEngine` did `context.LastNoun = ""` unconditionally, so `it` stopped resolving to an item the player was still holding.

## Fix

- Added `IContext.LastNouns` — the set of nouns the player last referred to as a group.
- Populate it from `take all` / `drop all`, multi-item take/drop, and several consecutive individual takes/drops.
- `ItProcessor` now resolves `them` against that set, expanding it into a conjoined noun list (`"sword and lantern"`) that the existing take/drop list handling already understands. The previous single-`IPluralNoun` behavior (candles) is preserved as a fallback.
- `MoveEngine` now only forgets antecedents for items left behind in the previous room; items still carried keep resolving via `it`/`them` after a move.

## Tests (TDD)

Added deterministic, no-AI tests that were red before the fix and green after:

- `ItProcessorTests`: `take all` → `drop them` drops everything taken; `take lantern; take sword; drop them` drops both.
- `PronounResolutionTests`: `it` survives a move for a carried item (`take lantern; examine it; east; drop it` → `Dropped`); an item left behind is correctly forgotten; updated the old test that asserted the buggy unconditional clear.

The default take/drop list-parser test mock now splits multi-item commands on `and`/commas, mirroring the real AI parser (existing multi-item tests that override the mock are unaffected).

### Verification

All affected suites pass locally:

| Suite | Result |
| --- | --- |
| UnitTests | 838 passed |
| ZorkOne.Tests | 1226 passed |
| Planetfall.Tests | 2232 passed |
| EscapeRoom.Tests | 7 passed |

(The `Lambda.Tests` project fails to NuGet-restore in this environment due to a pre-existing `Amazon.Lambda.Core` version conflict (NU1605), unrelated to these source-only changes.)

### Notes / out of scope

The issue also lists `take all. put them in the case`. `them` now *resolves* to the collection there too, but multi-item `put` is a separate engine limitation (`MultiNounEngine` handles a single direct object), so that command isn't fully wired up by this PR. The deterministic, high-value `drop them` / `take them` failures from the issue are fixed.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Jq5zWs11mtLpkSzYBWgtW4)_

### DIFF
--- a/GameEngine/Context.cs
+++ b/GameEngine/Context.cs
@@ -72,6 +72,8 @@ public abstract class Context<T> : IContext where T : IInfocomGame, new()
 
     public string LastNoun { get; set; } = "";
 
+    public List<string> LastNouns { get; set; } = new();
+
     public int Moves { get; set; }
 
     /// <summary>
--- a/GameEngine/Context.cs
+++ b/GameEngine/Context.cs
@@ -74,6 +74,24 @@ public abstract class Context<T> : IContext where T : IInfocomGame, new()
 
     public List<string> LastNouns { get; set; } = new();
 
+    public void RememberAntecedentNoun(string? noun)
+    {
+        if (!string.IsNullOrEmpty(noun) && !LastNouns.Contains(noun, StringComparer.OrdinalIgnoreCase))
+            LastNouns.Add(noun);
+    }
+
+    public void RememberAntecedentNouns(IEnumerable<string?> nouns)
+    {
+        var set = nouns
+            .Where(n => !string.IsNullOrEmpty(n))
+            .Select(n => n!)
+            .Distinct(StringComparer.OrdinalIgnoreCase)
+            .ToList();
+
+        if (set.Count > 0)
+            LastNouns = set;
+    }
+
     public int Moves { get; set; }
 
     /// <summary>
--- a/GameEngine/IntentEngine/MoveEngine.cs
+++ b/GameEngine/IntentEngine/MoveEngine.cs
@@ -37,7 +37,7 @@ public class MoveEngine : IIntentEngine
         // item after walking to the next room (issue #248).
         if (!context.HasMatchingNoun(context.LastNoun).HasItem)
             context.LastNoun = "";
-        context.LastNouns = (context.LastNouns ?? [])
+        context.LastNouns = context.LastNouns
             .Where(n => context.HasMatchingNoun(n).HasItem).ToList();
 
         return (null, await Go(context, generationClient, movement));
--- a/GameEngine/IntentEngine/MoveEngine.cs
+++ b/GameEngine/IntentEngine/MoveEngine.cs
@@ -32,9 +32,14 @@ public class MoveEngine : IIntentEngine
                 ? movement.CustomFailureMessage + Environment.NewLine
                 : await GetGeneratedCantGoThatWayResponse(generationClient, moveTo.Direction.ToString(), context));
 
-        // Let's reset the noun context, so we don't get confused with "it" between locations
-        context.LastNoun = "";
-        
+        // Forget pronoun antecedents for items we left behind in the previous room, but keep them
+        // for items the player is still carrying — "it"/"them" should still resolve to a carried
+        // item after walking to the next room (issue #248).
+        if (!context.HasMatchingNoun(context.LastNoun).HasItem)
+            context.LastNoun = "";
+        context.LastNouns = (context.LastNouns ?? [])
+            .Where(n => context.HasMatchingNoun(n).HasItem).ToList();
+
         return (null, await Go(context, generationClient, movement));
     }
 
--- a/GameEngine/IntentEngine/MultiNounEngine.cs
+++ b/GameEngine/IntentEngine/MultiNounEngine.cs
@@ -25,8 +25,10 @@ public class MultiNounEngine : IIntentEngine
         if (context.ItIsDarkHere)
             return (null, "It's too dark to see! ");
 
-        // After a multi-noun interaction, we will lose the ability to understand "it". 
+        // After a multi-noun interaction, we lose the ability to understand "it". It isn't a
+        // take/drop either, so it also ends any "them" group (issue #248).
         context.LastNoun = "";
+        context.LastNouns = [];
 
         var requireDisambiguation = CheckDisambiguation(interaction, context, interaction.MatchNounOne);
         if (requireDisambiguation is not null)
--- a/GameEngine/IntentEngine/SimpleInteractionEngine.cs
+++ b/GameEngine/IntentEngine/SimpleInteractionEngine.cs
@@ -1,4 +1,5 @@
 using System.Diagnostics;
+using Model;
 using Model.AIGeneration;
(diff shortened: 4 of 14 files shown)


arsindelve:
Thanks for the thorough review — addressed the real correctness gap in `f8a6fe8`:

- **Stale `LastNouns` (the main bug):** `them` now resolves only to remembered items still relevant to the command. A `drop` scopes to what is still **carried**, so an item dropped/eaten/left behind between takes no longer expands into `drop them` and triggers a spurious `DropSomethingTheyDoNotHave` narration. Other verbs (`take`/`put`/…) scope to reachable items (carried or present in the room). New red→green test `Them_DoesNotReferToItemsNoLongerInScope` covers exactly your scenario and asserts no AI call fires for the stale item. I went with verb-scoped resolution rather than resetting the group on every intervening command, since it fixes the same cases without changing `examine`/`look` UX (looking at something shouldn't forget what `them` means).
- **Dedup inconsistency:** batch take/drop antecedent sets are now de-duplicated case-insensitively, so two items sharing a primary noun can't produce `X and X`.
- **Lone-item fallback:** it now consults `LastNouns` before `LastNoun`.
- **Trailing newlines:** added to both batch processor files.

Consciously left as-is:
- **`DropIt` → `IDropSpecialLocation.DropSpecial` not recording an antecedent:** intentional — special-location drops (e.g. into the trophy case) are an uncommon flow and a low-value `them` antecedent; happy to add it if you'd prefer consistency.
- **Full consolidation of the `Remember*`/`SetAntecedent` helpers onto `Context`:** kept them local (with dedup now consistent) to avoid widening `IContext` for a minor; can pull them together if you'd rather.

---

arsindelve:
Good catch on the unbounded accumulation — fixed in `b79e75c`.

`LastNouns` now spans only a **contiguous run of take/drop commands**, enforcing the "consecutive" contract the PR claimed: `SimpleInteractionEngine` and `MultiNounEngine` clear it on any non-take/drop interaction (examine/open/put/…), mirroring how `LastNoun` is reset every turn. So your example now behaves:

```
take lantern        ; LastNouns = [lantern]
examine sword       ; group ends -> LastNouns = []
... 30 turns ...
take key            ; LastNouns = [key]
drop them           ; -> only the key is in the group (asks for clarification since a lone non-plural item isn't "them")
```

The lantern is no longer swept in. Genuinely consecutive takes (`take lantern. take sword. drop them`) still group as before. New red→green test `Them_DoesNotSpanUnrelatedInterveningCommands` covers it.

Also folded in the two nits from the review and a couple from a parallel pass:
- Centralized the take/drop verbs into `Model/Verbs.cs` (`TakeVerbs`/`DropVerbs`) and pointed `TakeOrDropInteractionProcessor`, `SimpleInteractionEngine`, and `ItProcessor` at them — no more private one-off `["drop"]` list.
- Consolidated the three duplicated antecedent helpers onto `Context` (`RememberAntecedentNoun`/`RememberAntecedentNouns`), case-insensitively de-duplicated — single source of truth.
- Removed the dead `?? []` guard in `MoveEngine` (the property is non-nullable) and made the unit-test mock reflect that.

All suites green: UnitTests 841, ZorkOne.Tests 1226, Planetfall.Tests 2232, EscapeRoom.Tests 7.

---

PR #265, opened 2026-06-22
Fix #262: "enter pod" (a valid noun) routes through the bulkhead instead of being mocked

### LINKED ISSUE #262: "enter pod" (a valid noun) is refused/mocked instead of giving "The escape pod bulkhead is closed."
## Bug

`enter pod` is a valid command — `BulkheadDoor.NounsForMatching` includes **`"pod"`** — but at Deck Nine it gets a generic refusal that the narrator dresses up as a mock of an imaginary object, instead of the correct **"The escape pod bulkhead is closed."**

Production (Deck Nine, fresh sessions):
- `enter pod` → *"Ah, trying to phase through solid metal? ..."*
- `board pod` → *"... teleports to 'podthat' are not yet installed ..."*
- `enter escape pod` → *"... that bulkhead remains firmly closed, Ensign."* ✅ (correct)

So the **full** phrase "escape pod" works; the bare valid noun **"pod"** does not.

## Root cause (deterministic, confirmed white-box — no AI)

`enter/board pod` is classified as an `EnterSubLocationIntent`. In `GameEngine/IntentEngine/EnterSubLocationEngine.cs` (lines 19-24):

```csharp
IItem? subLocation = Repository.GetItemInScope(enter.Noun, context);
if (subLocation == null)
    return (null, await GetGeneratedCantGoThatWayResponse(...));   // mock path
if (subLocation is not ISubLocation subLocationInstance)
    return (null, await GetGeneratedCantGoThatWayResponse(...));   // <-- "pod" lands here
```

At Deck Nine:
- `GetItemInScope("pod")` → **`BulkheadDoor`** (resolves fine — "pod" is a real noun)
- `BulkheadDoor is ISubLocation` → **False**
- → falls to `GetGeneratedCantGoThatWayResponse` → generic "cannot go that way" → narrator mock

The escape pod is reached by **going through the `BulkheadDoor`** (a door + movement), not by entering an `ISubLocation`. `EnterSubLocationEngine` has no path for "the noun is a door/openable" — it only enters sub-locations, otherwise it refuses generically. So a valid `enter <door>` is refused.

Reproduced deterministically:
```
StartHere<DeckNine>
GetItemInScope("pod")                       -> BulkheadDoor
BulkheadDoor is ISubLocation                -> False
EnterSubLocationEngine.Process(Noun="pod")  -> "You cannot go that way."  (generation disabled)
BulkheadDoor.IsOpen                         -> False  (so it should say "bulkhead is closed")
```

## Separate, secondary artifact: `podthat`

The `'podthat'` string in some responses is an AI-side noun-mangling artifact (the `<noun>` the AI emits for "board pod" comes through as `pod that`/`podthat`). It's a different layer — the routing bug above fires even with the clean noun `"pod"`, so this issue is about the routing. Noting `podthat` here for traceability.

## Expected
`enter pod` / `board pod` with the bulkhead shut → **"The escape pod bulkhead is closed."** (and once open, actually go through it), matching `enter escape pod`.

Found by the AdventureBreaker harness (black-box prod play → white-box trace). AB-018.

arsindelve:
## Problem

At Deck Nine, `enter pod` / `board pod` use a **valid** noun — `BulkheadDoor.NounsForMatching` includes `"pod"` — but the bare noun got a generic refusal that the narrator dressed up as a mock of an imaginary object ("Ah, trying to phase through solid metal? …"), instead of the correct **"The escape pod bulkhead is closed."** The full phrase `enter escape pod` worked because it's classified as a *Move*; the bare valid noun `pod` was not.

The issue's follow-up comment confirmed this is **general**, not pod-specific: `enter machine` at Physical Plant hit the same path.

## Root cause

`EnterSubLocationEngine` only knew how to enter `ISubLocation` items. When the noun resolved to a real, in-scope item that is a **door** (`IOpenAndClose`) rather than a sub-location, it fell through to `GetGeneratedCantGoThatWayResponse` → the narrator mock. The escape pod is reached by *moving through* the `BulkheadDoor`, so `enter <door>` needs to defer to movement — exactly as `enter escape pod` (a Move) already does.

## Fix

In `EnterSubLocationEngine`, when the resolved item is not an `ISubLocation`:

- **door (`IOpenAndClose`)** → defer to `MoveEngine` with `Direction.In`, so the location map applies its own open-check and custom failure message (`"The escape pod bulkhead is closed."`; once open, it actually carries you through).
- **anything else (machine, sword, …)** → return a plain `"You can't enter that."` instead of letting the narrator mock a perfectly valid object.

This same routing also repairs `enter window` at Zork's Behind House (the kitchen window is an `IOpenAndClose` door gating the `in` exit).

> Note: the secondary `podthat` noun-mangling artifact called out in the issue is a separate AI-layer concern and is intentionally **not** addressed here — this PR is the deterministic routing fix.

## Tests (TDD — red verified before the fix)

- **UnitTests / `EnterSubLocationEngineTests`** — a closed door defers to movement and yields the door's custom message (not "you cannot go that way"); a non-door in-scope item yields "You can't enter that."
- **ZorkOne.Tests / `KitchenWindowTests`** — `enter`/`board window` enters the kitchen when the window is open, and gives the closed-window message (staying put) when shut.
- **Planetfall.Tests / `EnterPodTests`** — `enter`/`board pod`, `enter bulkhead`, `enter door` all give "The escape pod bulkhead is closed." on Deck Nine; after the explosion opens the bulkhead, `enter`/`board pod` actually carries the player into the escape pod.

The TestParser intent mappings (`base-mappings.json`) were extended with the `enter pod`/`board pod`/`enter bulkhead`/`enter door`/`enter window` phrasings so the full-engine tests parse these deterministically, mirroring how the production AI parser classifies them.

## Verification

All three affected suites pass with no regressions:

- `UnitTests` — 843 passed
- `ZorkOne.Tests` — 1230 passed
- `Planetfall.Tests` — 2252 passed

---
_Generated by [Claude Code](https://claude.ai/code/session_01CrxNqNsaW6EjQ1eV9Y4jPE)_

### DIFF
--- a/GameEngine/IntentEngine/DoorReroute.cs
+++ b/GameEngine/IntentEngine/DoorReroute.cs
@@ -2,50 +2,43 @@ using Model.AIGeneration;
 using Model.Intent;
 using Model.Interface;
 using Model.Item;
-using Model.Movement;
 
 namespace GameEngine.IntentEngine;
 
 /// <summary>
-/// Shared "enter/exit &lt;door&gt;" routing for <see cref="EnterSubLocationEngine"/> and
-/// <see cref="ExitSubLocationEngine"/> (issue #262). Kept in one place so the two engines can't
-/// drift apart on the (subtle) definition of a door.
+/// Shared "enter/exit/go-through &lt;door&gt;" routing for <see cref="EnterSubLocationEngine"/> and
+/// <see cref="ExitSubLocationEngine"/> (issue #262).
 ///
-/// A <b>passage door</b> is a fixed openable that gates a room exit: it is
-/// <see cref="IOpenAndClose"/>, NOT something you carry (<see cref="ICanBeTakenAndDropped"/>), and
-/// NOT a container (<see cref="ICanContainItems"/>). The container exclusion matters because
-/// <see cref="Repository.GetItemInScope"/> also searches inventory and the room, so without it an
-/// openable that is merely <i>present</i> — a carried sack, or the Living Room's wall-mounted trophy
-/// case — would resolve here and hijack the room's exit (e.g. "enter case" teleporting you down the
-/// trap door). "enter/exit &lt;door&gt;" means "go through it", so we defer to <see cref="MoveEngine"/>
-/// in the given direction and let the location map's own open-check / CustomFailureMessage apply,
-/// instead of a generic narrator-mock refusal.
+/// A door is whatever a room's map <i>declares</i> as the <see cref="Model.Movement.MovementParameters.GatingItem"/>
+/// of one of its exits — a window, a bulkhead, a trap door, a grating. We do NOT guess doorness from
+/// the item's type; we ask the current location which direction the resolved noun gates
+/// (<see cref="Model.Location.ILocation.DirectionGatedBy"/>) and walk that way, letting the map's own
+/// open-check / CustomFailureMessage apply. Because the map names the door, a carried sack or a
+/// wall-mounted trophy case is simply not anyone's gating item and falls through to the caller's
+/// refusal — no portability/container exclusions needed.
 ///
-/// NOTE: this confirms the noun is <i>a</i> passage door and that the room has <i>that</i> exit, not
-/// that <i>this</i> door gates <i>that</i> exit. It is safe as long as no room we expose an In/Out
-/// door-alias in has two distinct passage doors sharing a noun. (BioLabLocation has two "door"s, but
-/// it is deliberately NOT aliased.) Tying a door to its gating direction would remove that caveat.
+/// From a given room a door gates exactly one passage, so "enter", "exit" and "go through" all mean
+/// the same thing here: traverse it. There is deliberately no In/Out direction — the map says which
+/// way the door leads.
 /// </summary>
 internal static class DoorReroute
 {
     /// <summary>
-    /// If <paramref name="resolvedNoun"/> is a passage door and the current location has an exit in
-    /// <paramref name="direction"/>, walk through it via <see cref="MoveEngine"/> and return the
-    /// result. Otherwise returns null, leaving the caller to give its own "you can't go through that"
-    /// message.
+    /// If <paramref name="resolvedNoun"/> is the item gating one of the current location's exits,
+    /// walk through it via <see cref="MoveEngine"/> and return the result. Otherwise returns null,
+    /// leaving the caller to give its own "you can't go through that" message.
     /// </summary>
-    public static async Task<(InteractionResult? resultObject, string ResultMessage)?> TryProcess(
-        IItem? resolvedNoun, Direction direction, IContext context, IGenerationClient generationClient)
+    public static async Task<(InteractionResult? resultObject, string ResultMessage)?> TryTraverse(
+        IItem? resolvedNoun, IContext context, IGenerationClient generationClient)
     {
-        if (resolvedNoun is not IOpenAndClose
-            || resolvedNoun is ICanBeTakenAndDropped
-            || resolvedNoun is ICanContainItems)
+        if (resolvedNoun is null)
             return null;
 
-        if (context.CurrentLocation.Navigate(direction, context) is null)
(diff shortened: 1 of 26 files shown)


arsindelve:
Thanks — good catch, and the facts check out: `OfficeDoor` is gated on `Direction.W` and `TrapDoor` on `Direction.Up`, neither has an `In` entry, so the unconditional `Direction.In` reroute wouldn't reach them.

One clarification: it isn't a *regression*. Before this PR, **every** non-sub-location `enter <item>` already fell through to the generic refusal/mock — this PR fixes the `In`-gated subset (the reported pod bug) and leaves the rest as they were. So nothing gets worse; it's an incomplete generalization.

Addressed in `ab58ab2`: the reroute is now **guarded** to only defer to movement when the current location actually gates an `In` passage (`Navigate(Direction.In) is not null`, a pure map lookup). When there's no `In` exit, we return the plain `"You can't enter that."` instead of the generic "you cannot go that way" mock. This keeps the full #262 fix for the pod + kitchen window while no longer routing `OfficeDoor`/`TrapDoor`/etc. into the generic refusal. Added an engine test using the Living Room trap door (reached via "down", no "in" exit) to lock it in.

The *complete* generalization — `enter <any door>` routing to whichever direction that door gates — needs a door→direction link the movement model doesn't expose today (`Map()` gates via opaque `CanGo` lambdas with no item→direction back-reference). I've filed that as a follow-up: #266.

Re: the minor note on `LastMovementDirection` now being reachable from a door-noun "enter" — agreed it's semantically defensible (the player did go "in") and intended.

---

arsindelve:
Thanks — all three addressed in `75fa7a2`:

1. **Exit-side symmetry.** `ExitSubLocationEngine` now returns a plain `"You can't exit that. "` when the resolved noun is neither a sub-location nor a door, mirroring `"You can't enter that. "` — no more narrator-mock for `exit sword`. Updated the exit-engine unit test to match.
2. **Latent two-openables trap.** Added a `NOTE` comment to both engines spelling out that the guard checks for *a* fixed door + *an* in/out exit, not that this specific door gates it — safe today, and the general case is tracked in #266.
3. **DRY.** `LivingRoom`/`Cellar` now build the trap-door `MovementParameters` once and share it across the `down`/`up` and `in` keys, matching `BehindHouse`'s existing `enterKitchen` pattern.

Also (separate maintainer request) verified `exit pod` once the pod has landed in the ocean: bulkhead open → carries you out to `Underwater` (the intended escape), closed → "The pod door is closed." Added `ExitPodUnderwater` regression tests. All suites green: UnitTests 844, ZorkOne.Tests 1235, Planetfall.Tests 2254.

---

arsindelve:
Great review — the "two distinct fixed openables" claim being already false turned out to matter more than harmless. Fixed in `7de2822`:

- **It was a live bug, not just a stale comment.** The Living Room is itself a two-openable room — the trap door (now a `Direction.In` passage) **and** the wall-mounted trophy case. So `enter case` resolved the trophy case, passed the old guard (non-portable `IOpenAndClose`), and teleported the player down the trap door into the Cellar. Reproduced with a red test, then fixed: a **passage door is an openable that is neither portable nor a container** (`is not ICanContainItems`). Trophy case / mailbox / sack are `OpenAndCloseContainerBase ⇒ ICanContainItems` and are now excluded; the real doors (`ItemBase, IOpenAndClose`) aren't.
- **De-duplicated.** The guard + rationale are now a single `DoorReroute.TryProcess` helper both engines call, so the door definition can't drift — and the comment states the *accurate* invariant (no `In`/`Out`-aliased room may have two same-noun passage doors; `BioLabLocation` is named as the case to avoid).
- **`LabOffice` gap:** good catch — deliberately deferred (it leads to the `BioLabLocation` fungicide death-check room, so `enter door` would be a death synonym for `W`), and `BioLabLocation` itself is the concrete two-`"door"` ambiguity. Both tracked in reopened #266.
- Minor (double `Navigate`, inline `new MoveEngine()`): acknowledged, left as-is per the existing pattern.

Suites green: UnitTests 844, ZorkOne.Tests 1238, Planetfall.Tests 2260.

---

PR #267, opened 2026-06-22
Fix #264: addressing an absent talkable NPC no longer leaks into player actions

### LINKED ISSUE #264: Addressing an absent talkable NPC (Floyd/Blather/Ambassador) leaks into player actions instead of saying "X isn't here"
## Bug

The game only "knows" Floyd / Blather / the Ambassador when they're physically present. When you address one **while they're absent**, the direct-address isn't recognized at all and the input falls through to normal parsing — so the **player executes the command**, including real state changes (movement, dropping items), or the narrator **hallucinates the NPC acting**.

Expected: the game always knows these named characters, and addressing one who isn't present should say so — e.g. *"Floyd isn't here."* — and **never** move the player, change inventory, or pretend the NPC acted.

This is the absent-case gap left by #182 (which fixed direct-address routing **when the character is present**).

## Reproductions (production, fresh sessions at Deck Nine — all three NPCs absent)

| Command | Actual | Should be |
|---|---|---|
| `floyd, take brush` | "You already have that!" (player) | "Floyd isn't here." |
| `floyd, drop diary` | "Dropped" — **player inv 4→3** | "Floyd isn't here." |
| `floyd, go up` | **player walks to Gangway** | "Floyd isn't here." |
| `floyd, sing` | "Floyd… belts out a tune" — **hallucinates Floyd present** | "Floyd isn't here." |
| `blather, drop diary` | "Dropped" — player drops it | "Blather isn't here." |
| `blather, go up` | **player walks to Gangway** | "Blather isn't here." |
| `ambassador, take brush` | "You already have that!" (player) | "The ambassador isn't here." |
| `ambassador, go up` | narrator: "the elusive ambassador… not here" (no move) | "The ambassador isn't here." |

`floyd, hello` and `ambassador, go up` occasionally produce a correct "not here" narration, so the behavior is **inconsistent** — it depends on whether the fall-through command parse happens to do something.

The movement/drop cases are the worst: a player who fat-fingers a Floyd command when Floyd has wandered off can be **silently walked into another room or made to drop a needed item**.

## Root cause (white-box)

`GameEngine/ConversationHandler.cs`, `CollectTalkableEntities` (lines ~53-65) only gathers `ICanBeTalkedTo` characters from the player's inventory and the **current location**:

```csharp
var talkers = new List<ICanBeTalkedTo>();
talkers.AddRange(context.Items.OfType<ICanBeTalkedTo>());            // inventory
if (context.CurrentLocation is ICanContainItems container)
    talkers.AddRange(container.Items.OfType<ICanBeTalkedTo>());      // current room only
```

If the addressed NPC isn't in that set, `FindTargetCharacter` returns null → `CheckForConversation` returns null → the input goes to normal command parsing, where the leading `"name,"` is effectively ignored and the remainder runs as a **player** command.

## Suggested direction (not prescribing the fix)

Recognize direct address to **any known talkable NPC by name/synonym regardless of location** — e.g. a global registry of `ICanBeTalkedTo` characters and their `NounsForMatching` — and when the addressed character is recognized but **not in scope**, short-circuit with "X isn't here." rather than falling through to player parsing. (Per the owner: a new classifier prompt for direct-address-to-known-character can be added if needed.)

Affects all three `ICanBeTalkedTo` NPCs (Floyd, Blather, Ambassador — marked in #154).

Found by the AdventureBreaker harness playing production. AB-019.

arsindelve:
## Summary

Fixes #264. Addressing Floyd / Blather / the Ambassador by name **while they are absent** used to fall through to normal command parsing, so the **player** executed the command — silently moving rooms, dropping items, or letting the narrator hallucinate the NPC acting. This is the absent-case gap left by #182 (which fixed direct address only when the character is present).

Now the engine always *knows* its talkable characters, and directly addressing one who isn't present short-circuits with a terse, deterministic **"X isn't here."** instead of leaking the remainder into player parsing.

| Command (NPC absent) | Before | After |
|---|---|---|
| `floyd, go up` | player walks to Gangway | "Floyd isn't here." (no move) |
| `floyd, drop diary` | player drops it (inv 4→3) | "Floyd isn't here." (kept) |
| `blather, take brush` | "You already have that!" | "Blather isn't here." |
| `ambassador, go up` | inconsistent narration | "The ambassador isn't here." |
| `tell floyd to go up` | leaks | "Floyd isn't here." |

## How it works

- **`IInfocomGame.TalkableCharacterTypes`** — each game declares its talkable roster (default: none). `PlanetfallGame` lists `Floyd`, `Blather`, `Ambassador`, so they're "known" regardless of location and *even before* lazy instantiation (the worst reproduction, `floyd, go up` on Deck Nine, happens before Floyd is ever loaded).
- **`ConversationHandler`** resolves that roster when no talker is present, and only fires when the input is a *genuine direct address*:
  - vocative `Name, …` (reuses the existing `TryStripDirectAddress`), and
  - imperative `tell/ask/talk to/yell at … Name …`, but only when the name immediately follows the address verb.
  - Conservative by design: real commands that merely mention the name (`examine floyd`, `attack floyd`, `ask about the alien`) are **not** hijacked and fall through to normal parsing.
  - The guard is **deterministic — no AI** — so it runs even in `NoGeneratedResponses` mode and is fully unit-testable.
- **`ICanBeTalkedTo.NotHereDescription`** supplies the static reply (capitalized matching name by default; `Ambassador` overrides to `"The ambassador isn't here. "` since "ambassador" is a title).
- **`Repository.GetItem(Type)`** added to resolve roster entries by runtime type (mirrors `GetItem<T>()`).
- Present-NPC conversation routing (#182) is **unchanged**.

## Tests (TDD — red before, green after)

- `UnitTests/Engine/ConversationHandlerTests` — handler-level, deterministic, with mocked seams: absent vocative + imperative address → "isn't here", runs when generation disabled, overridden phrasing, present-NPC precedence, and several no-hijack cases.
- `Planetfall.Tests/AbsentTalkableNpcTests` — full-engine with the real NPCs, asserting both the message **and** that the player did not move / did not drop the item.

All suites green: `UnitTests` (856), `Planetfall.Tests` (2258), `ZorkOne.Tests` (1226), `EscapeRoom.Tests` (7); full solution builds with 0 warnings/0 errors.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HzzV8FfuhzoBya2uf3MYrC)_

### DIFF
--- a/GameEngine/ConversationHandler.cs
+++ b/GameEngine/ConversationHandler.cs
@@ -1,6 +1,7 @@
 using ChatLambda;
 using Microsoft.Extensions.Logging;
 using Model.AIGeneration;
+using Model.AIGeneration.Requests;
 using Model.Interface;
 using Model.Item;
 
@@ -31,19 +32,22 @@ public class ConversationHandler(
         if (targetCharacter == null)
         {
             // The addressed character (if any) is not in scope. If the player nonetheless directly
-            // addressed a KNOWN talkable character by name — the vocative "Name, ..." form — answer
-            // "X isn't here." instead of letting the leftover command fall through to normal player
-            // parsing. Falling through is the #264 bug: it would silently move the player
-            // ("floyd, go up"), drop their items ("floyd, drop diary"), or let the narrator
-            // hallucinate the absent NPC acting. This guard is deterministic (no AI), so it runs
-            // even when generation is disabled and is fully unit-testable.
-            var absentTarget = FindTargetCharacter(input, CollectAllKnownTalkers());
-            if (absentTarget != null && IsGenuineDirectAddress(input, absentTarget))
-            {
-                var notHere = absentTarget.NotHereDescription;
-                logger?.LogDebug($"[CONVERSATION DEBUG] Addressed absent talker; responding: '{notHere}'");
-                return notHere;
-            }
+            // addressed a KNOWN talkable character by name — the vocative "Name, ..." or imperative
+            // "tell/ask ... Name" form — tell them the character isn't here instead of letting the
+            // leftover command fall through to normal player parsing. Falling through is the #264
+            // bug: it would silently move the player ("floyd, go up"), drop their items ("floyd, drop
+            // diary"), or let the narrator hallucinate the absent NPC acting. The *detection* here is
+            // deterministic (no AI), so it always fires — even when generation is disabled — and is
+            // fully unit-testable; the response itself is narrated (see NarrateAbsence).
+            //
+            // Scan every known talker (not just the first whose name appears) so that when a
+            // sentence mentions more than one of them, the one actually being addressed wins.
+            // IsGenuineDirectAddress is a stricter check than FindTargetCharacter's substring match,
+            // so it both identifies and validates the addressed character on its own.
+            var absentTarget = CollectAllKnownTalkers()
+                .FirstOrDefault(talker => IsGenuineDirectAddress(input, talker));
+            if (absentTarget != null)
+                return await NarrateAbsence(absentTarget, context);
 
             logger?.LogDebug("[CONVERSATION DEBUG] No matching character found in input, returning null");
             return null;
@@ -58,6 +62,33 @@ public class ConversationHandler(
         }
 
         return await ProcessConversation(input, targetCharacter, context);
+    }
+
+    /// <summary>
+    /// Produces the response when the player addressed an absent known character. The narrator tells
+    /// them the character isn't here in its own voice (the owner wants this generated, not a fixed
+    /// line). The static <see cref="ICanBeTalkedTo.NotHereDescription"/> is used only as a fallback
+    /// when generation is unavailable (NoGeneratedResponses mode) or returns nothing — so the guard
+    /// is still deterministic in those modes and never leaks back into player parsing.
+    /// </summary>
+    private async Task<string> NarrateAbsence(ICanBeTalkedTo target, IContext context)
+    {
+        var fallback = target.NotHereDescription;
+
+        if (generationClient.IsDisabled)
+        {
+            logger?.LogDebug($"[CONVERSATION DEBUG] Generation disabled; absent talker fallback: '{fallback}'");
+            return fallback;
+        }
+
+        var characterName = (target as IItem)?.Name ?? "that character";
+        var request = new TalkingToAbsentCharacterRequest(
+            context.CurrentLocation.GetDescriptionForGeneration(context), characterName);
+
+        var narration = await generationClient.GenerateNarration(request, context.SystemPromptAddendum);
+        logger?.LogDebug($"[CONVERSATION DEBUG] Narrated absent talker '{characterName}': '{narration}'");
+
+        return string.IsNullOrWhiteSpace(narration) ? fallback : narration;
     }
 
     /// <summary>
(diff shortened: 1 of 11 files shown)


GameEngine/ConversationHandler.cs
```diff
@@ -22,29 +23,121 @@ public class ConversationHandler(
...
+    ];
+
+    /// <summary>
+    /// A conservative, deterministic test for whether the player genuinely addressed the character.
+    /// Catches the vocative "Name, ..." form and the imperative "tell/ask/talk to ... Name ..."
+    /// forms, but only when the character's name appears immediately after the address verb. Real
+    /// commands that merely mention the name ("examine floyd", "look at the robot", "ask about the
+    /// alien") are NOT hijacked and correctly fall through to normal parsing.
+    /// </summary>
+    private static bool IsGenuineDirectAddress(string input, ICanBeTalkedTo target)
```
arsindelve, line 200: **[High] The fix is incomplete — bare no-comma direct address still leaks.**

`IsGenuineDirectAddress` only fires on a comma (via `TryStripDirectAddress`) or a lead-in verb (`tell/ask/talk to/…`). A bare `floyd go up` / `floyd drop diary` matches neither, so it falls through to normal parsing — the exact #264 harm.

Verified against **current prod** (which lacks this fix):
- `floyd go up` → player walks to **Gangway**
- `floyd drop diary` → "Dropped", inventory **4→3**

Since this branch's guard provably won't catch those (no comma, no lead-in), the bare vocative still moves the player / drops items. The PR title says "no longer leaks into player actions," but it only closes the comma'd and lead-in forms.

Suggestion: also treat a **leading NPC name followed by more words** (`Name <rest>` at position 0, even without a comma) as direct address — almost nothing legitimately starts a command with "floyd"/"blather". At minimum, add a test pinning the intended behavior of `floyd go up` so the scope decision is explicit.

---

GameEngine/ConversationHandler.cs
```diff
@@ -22,29 +23,121 @@ public class ConversationHandler(
...
+        if (targetCharacter == null)
+        {
+            // The addressed character (if any) is not in scope. If the player nonetheless directly
+            // addressed a KNOWN talkable character by name — the vocative "Name, ..." form — answer
+            // "X isn't here." instead of letting the leftover command fall through to normal player
+            // parsing. Falling through is the #264 bug: it would silently move the player
+            // ("floyd, go up"), drop their items ("floyd, drop diary"), or let the narrator
+            // hallucinate the absent NPC acting. This guard is deterministic (no AI), so it runs
+            // even when generation is disabled and is fully unit-testable.
+            var absentTarget = FindTargetCharacter(input, CollectAllKnownTalkers());
```
arsindelve: **[Low] Coarse `Contains` matching can mis-attribute the absent NPC.**

`FindTargetCharacter` matches any input that *contains* a noun, and Floyd's nouns include `"robot"`. So `tell the robot to …` — aimed at one of the other (non-talkable) robots, e.g. in the Robot Shop — while Floyd is absent will resolve to Floyd and answer *"Floyd isn't here."*

Low impact (Floyd is "the" robot companion), but it's a false attribution worth noting. If you want to tighten it, gate the absent-path match on a word-boundary check like the `StartsWithWholeWord` helper you already added, rather than substring `Contains`.

---

arsindelve (commented):
Solid, well-structured fix for #264 — game-declared roster, deterministic absent-address guard, good test coverage. I verified `Repository.GetItem(Type)` is consistent with `GetItem<T>()` (same instantiate→Init→cache singleton, so the guard can't corrupt/duplicate NPC state) and that the no-hijack cases hold.

One blocking-ish gap (inline): the guard is comma/lead-in gated, so the **bare no-comma form `floyd go up` still leaks in prod** (verified — player moves / drops items). I'd hold the "fixed" claim until that's handled or explicitly scoped out. Plus one low-severity coarse-match note.

---

arsindelve:
Pushed `7861c20` addressing the review:

- **[High] bare no-comma leak** — a leading character **name** now counts as direct address even without a comma or lead-in verb, so `floyd go up` / `blather drop diary` are caught instead of leaking to player parsing. Pinned by new tests (`floyd go up`, `floyd drop diary`, `blather go up`).
- **[Low] generic synonym mis-attribution** — absent-address matching is now restricted to the character's actual **names** (primary name + title-prefixed variants like `ensign blather`) and excludes generic synonyms like `robot`/`alien`, with word-boundary matching. So `tell the robot to …` aimed at another robot is no longer answered as Floyd (test added). The player still addresses Floyd as `floyd`.

Also, per the request to not use a canned line: the absent reply is now **narrated** (`GenerateNarration` + a new `TalkingToAbsentCharacterRequest`); the static `NotHereDescription` is only a deterministic fallback for `NoGeneratedResponses`/empty-generation, so detection stays deterministic and never leaks.

Detection scans every known talker so the genuinely-addressed one wins when a sentence names more than one. Full suites green locally (UnitTests, Planetfall.Tests, ZorkOne.Tests, EscapeRoom.Tests).

---

arsindelve:
Pushed `5f1edf5` (rebased on the latest `main` merge) with a reworked absent-address path, per owner decisions in this session:

- **LLM-backed detection**, symmetric with the present-NPC path: a deterministic fast path catches the explicit forms (and works offline), and anything else defers to the same `ParseConversation` classifier — so arbitrary phrasing ("could you let floyd know to wait") is recognized. Classifier is skipped when generation is disabled, keeping `NoGeneratedResponses` deterministic.
- **Generic synonyms now intentionally count** (Floyd answers to "robot", ambassador to "alien"). This deliberately reverses the earlier "exclude synonyms" note — the owner confirmed Floyd IS the robot and no other robot in Planetfall is talkable. So `the robot, go up` / `hey robot` reach Floyd.
- **Casual openers** ("hey/yo/hi/hello") and the **leading-article** form ("the ambassador, go up") are recognized.

Detection still identifies the target by a whole-word name/synonym match; the reply is narrated with the static line as the offline/empty fallback. New/updated tests cover all of the above plus the offline (no-classifier) behavior. Local suites green: UnitTests 877, Planetfall.Tests 2267, ZorkOne.Tests 1226.

---

Issue #256, opened 2026-06-22
Production: HTTP 500 on multi-clause input "look examine bulkhead open bulkhead"


arsindelve:
## Bug

Against **Planetfall production** (`https://6kvs9n5pj4.execute-api.us-east-1.amazonaws.com/Prod/Planetfall`), the single input:

```
look examine bulkhead open bulkhead
```

returns **HTTP 500** (empty body) — reproducible from a **fresh session at move 0**. The server should return a graceful parse/"I didn't understand" response, not a 500.

## Isolation (fresh session per input, to rule out game-state contamination)

| Input | Status |
|---|---|
| `look examine bulkhead open bulkhead` | **500** |
| `examine bulkhead open bulkhead` | 200 |
| `look open bulkhead` | 200 |
| `examine open bulkhead` | 200 |
| `look examine bulkhead open` | 200 |
| `x bulkhead open bulkhead` | 200 |
| `look examine examine examine` | 200 |
| `open bulkhead` | 200 |

So it's not a single verb/noun — it's the specific 3-clause shape `look` + `examine <noun>` + `open <noun>`. Likely an unhandled exception in the multi-clause / sentence-splitting parse path (no periods in the input).

## Repro

```
GET  /Planetfall?sessionId=<NEW_GUID>
POST /Planetfall  {"input":"look examine bulkhead open bulkhead","sessionId":"<NEW_GUID>"}
  -> HTTP 500, no body
```

A 500 is a hard failure (no narrative, transport error to the client) rather than the engine's normal graceful degradation, so flagging as high.

Found by the AdventureBreaker harness playing production (black-box). Logged as AB-015.

Next step: reproduce against the engine in a unit test to capture the throwing code path, then fix + harden so malformed multi-clause input degrades gracefully.

arsindelve:
## Further characterization (production, fresh sessions)

The 500 is **extremely specific** — it requires the exact 4-clause shape `look` + `examine bulkhead` + `open bulkhead`, with the nouns resolving to a **real in-scope object**. Changing any single element makes it a graceful 200:

| Input (fresh session) | Status |
|---|---|
| `look examine bulkhead open bulkhead` | **500** |
| `examine bulkhead open bulkhead` (drop leading `look`) | 200 |
| `look open bulkhead open bulkhead` (1st verb → open) | 200 |
| `look examine bulkhead examine bulkhead` (2nd verb → examine) | 200 |
[...]

PR #274, opened 2026-06-22
Add #271: engine safety net — never return HTTP 500, narrator "oops" on any unhandled crash

### LINKED ISSUE #271: Engine safety net: never return HTTP 500 — narrator "oops" on any unhandled crash
## Summary

No matter what goes wrong inside the engine, the player should **never** see a raw HTTP 500 / empty body. Any unhandled exception during turn processing should be caught at a single chokepoint, logged, and converted into a graceful, in-character narrator **"oops, something went wrong"** response returned with **HTTP 200**.

This is a **defense-in-depth safety net**, complementary to (not a replacement for) fixing the specific bugs that throw. Individual crashes (e.g. #256, the duplicate-tag parser throw) still get proper root-cause fixes; this guarantees that the *next* unknown bug degrades gracefully instead of breaking the transport to the client.

## Motivation

#256 is a concrete example: a parser edge case threw `InvalidOperationException`, propagated uncaught through the Lambda controller, and surfaced as an HTTP 500 with no body — a hard failure with no narrative, rather than the engine's usual graceful degradation. The class of "some unforeseen input throws somewhere deep in the engine" is open-ended; we want a structural guarantee that it can never again reach the player as a 500.

## Desired behavior

- Any unhandled exception thrown while producing a turn response is caught.
- The player receives a short, **AI-generated narrator** message in the game's voice — something like *"Something goes wrong, and for a moment the world seems to flicker. Perhaps try that a different way."* — **not** a hardcoded string (consistent with how the engine already generates deflection/no-op narration).
- Response status is **200**, with the normal `GameResponse` shape (so the web/mobile clients render it like any other turn).
- The exception is logged (CloudWatch, with the turn correlation id) so we can still find and fix the underlying bug.
- Session state is not corrupted; the player can keep playing.

## Proposed implementation

**Primary chokepoint — `GameEngine.GetResponse` (covers every game, every turn).**
Wrap the body of `GameEngine/GameEngine.cs:GetResponse(...)` in a try/catch. On exception:
1. Log the exception with `_turnCorrelationId` via the existing CloudWatch logger.
2. Generate narration from a new `EngineErrorRequest` (below) and return it through `PostProcessing` so it behaves like a normal turn.
3. If generation itself is unavailable/throws (e.g. the AI client is the thing that failed, or `NoGeneratedResponses` is set), fall back to a **single guaranteed static** sentence so we still return 200. (The static fallback is the one acceptable canned string — it exists only for the case where the narrator itself can't be reached.)

Because all turn entry points (`POST /`, the `GET /` look-on-reconnect, and the `restoreGame` "look") funnel through `GetResponse`, this one wrap covers all of them across Zork, Planetfall, and EscapeRoom.

**New request type — `Model/AIGeneration/Requests/EngineErrorRequest.cs`.**
Follows the existing deflection pattern (`CommandHasNoEffectOperationRequest`): low `DeflectionTemperature`, includes `NoInventionGuard`, and asks the narrator to acknowledge — briefly and in-character — that something went wrong and the action couldn't be completed, without inventing world state and without revealing the error.

**Secondary chokepoint (optional, broader) — ASP.NET exception middleware.**
The save/restore/session endpoints don't go through `GetResponse` and can also throw (Base64 decode, JSON deserialize, DynamoDB). Add exception-handling middleware in the shared server/`Startup` so those endpoints also return a graceful `GameResponse`/error payload instead of a 500. Lower priority than the turn-level net, but closes the remaining gap.

## Testing (deterministic — no AI, no network)

- Engine-level: construct a `GameEngine` with a mocked parser/processor seam that **throws** for a given input; assert `GetResponse(...)` does **not** throw, returns a non-empty 200-style string, and that the mocked `IGenerationClient.GenerateNarration` was invoked with an `EngineErrorRequest`.
- Fallback path: mock `IGenerationClient.GenerateNarration` to also throw; assert `GetResponse(...)` still returns the guaranteed static fallback and does not rethrow.
- `RequestTests`: assert `EngineErrorRequest.UserMessage` is non-empty and uses the deflection temperature.

## Scope

- **In scope:** a single, well-placed catch-all that converts engine turn crashes into a 200 narrator response across all games; structured logging of the swallowed exception; deterministic tests.
- **Out of scope:** fixing any specific underlying bug (those get their own issues/PRs, e.g. #256). The safety net must not *hide* bugs — it must log loudly enough that they're still discoverable.

## Acceptance criteria

- [ ] An exception thrown anywhere during turn processing yields a 200 narrator "oops" response, never a 500.
- [ ] The narrator message is AI-generated (with a guaranteed static fallback only when generation is itself unavailable).
- [ ] The exception is logged with the turn correlation id.
- [ ] Behavior verified by deterministic unit tests.
- [ ] Applies to Zork, Planetfall, and EscapeRoom (shared engine path).

---
_Filed alongside #256 (a specific instance of the failure this net would have caught)._

arsindelve:
Closes #271.

## What

A single, well-placed catch-all so that **no unhandled exception during turn processing can ever reach the player as a raw HTTP 500 / empty body**. Any crash deep in the engine is now caught, logged, and converted into a graceful, in-character narrator **"oops"** returned with the normal `GameResponse` shape (a non-empty 200 body).

This is **defense-in-depth**, complementary to — not a replacement for — fixing the specific bugs that throw (e.g. #256). The net deliberately logs *loudly* so the underlying bug stays discoverable.

## How

**Primary chokepoint — `GameEngine.GetResponse`.** The body is wrapped in a single `try/catch`. Because every turn entry point (`POST /`, the `GET /` reconnect "look", and the `restoreGame` "look") funnels through `GetResponse`, this one wrap covers **Zork, Planetfall, and EscapeRoom**.

On an unhandled exception, `HandleUnexpectedEngineError`:
1. Logs the exception via the engine logger **and**, best-effort, writes a structured CloudWatch `TurnLog` carrying the **turn correlation id** — so the swallowed exception is still findable.
2. Generates a short, in-character acknowledgement from a new **`EngineErrorRequest`** and returns it through the normal `PostProcessing` pipeline, so it renders like any other turn.
3. Falls back to a **single guaranteed static sentence** only when the narrator itself is unavailable (`NoGeneratedResponses` set, or generation itself throws). `SafePostProcess` guarantees a non-empty 200 body even if post-processing itself throws.

**New request type — `Model/AIGeneration/Requests/EngineErrorRequest.cs`.** Follows the existing deflection pattern (`CommandHasNoEffectOperationRequest`): low `DeflectionTemperature`, includes `NoInventionGuard`, asks the narrator to acknowledge — briefly and in-character — that the action couldn't be completed, **without** inventing world state and **without** revealing the error.

## Testing (deterministic — no AI, no network)

- `GenerationFallbackTests.UnhandledException_DuringTurn_ReturnsGeneratedNarration_WithoutThrowing` — a parser seam that throws yields a non-empty, non-throwing response, and `IGenerationClient.GenerateNarration` is invoked with an `EngineErrorRequest`.
- `GenerationFallbackTests.UnhandledException_WhenGenerationAlsoThrows_ReturnsStaticFallback` — when error-narration generation *also* throws, the guaranteed static fallback is returned and nothing rethrows.
- `RequestTests.EngineErrorRequest_HasNonEmptyMessage_AndUsesDeflectionTemperature` — message is non-empty and runs at the low deflection temperature.

Confirmed red first (temporarily neutralised the catch with an always-false exception filter → both engine tests failed on `NotThrowAsync` because the exception propagated), then green with the fix.

Full suites pass with no regressions: UnitTests (880), ZorkOne.Tests (1228), Planetfall.Tests (2267), EscapeRoom.Tests (7).

## Acceptance criteria

- [x] An exception thrown anywhere during turn processing yields a 200 narrator "oops" response, never a 500.
- [x] The narrator message is AI-generated (with a guaranteed static fallback only when generation is itself unavailable).
- [x] The exception is logged with the turn correlation id.
- [x] Behavior verified by deterministic unit tests.
- [x] Applies to Zork, Planetfall, and EscapeRoom (shared engine path).

## Out of scope

The issue's **optional secondary chokepoint** (ASP.NET exception middleware for the save/restore/session endpoints that don't go through `GetResponse`) is intentionally left as a follow-up. It's explicitly lower priority, touches the shared hosting layer, and the turn-level net already satisfies every acceptance criterion above.

---
_Generated by [Claude Code](https://claude.ai/code/session_01S4DxLeBTKeEbeszUPzdrU4)_

### DIFF
--- a/GameEngine/GameEngine.cs
+++ b/GameEngine/GameEngine.cs
@@ -225,6 +225,12 @@ public class GameEngine<TInfocomGame, TContext> : IGameEngine
         }
         catch (Exception ex)
         {
+            // Intentionally broad — this deliberately includes OperationCanceledException /
+            // TaskCanceledException, which is also how HttpClient timeouts (e.g. the OpenAI
+            // generation call) surface. Those are genuine "something went wrong deep in the engine"
+            // failures that must degrade gracefully. GetResponse has no CancellationToken to tell a
+            // real client-cancel apart from a timeout, so filtering cancellation out here would let
+            // timeouts regress straight back into the HTTP 500s this net exists to prevent.
             return await HandleUnexpectedEngineError(ex);
         }
     }
--- a/GameEngine/GameEngine.cs
+++ b/GameEngine/GameEngine.cs
@@ -55,6 +55,15 @@ public class GameEngine<TInfocomGame, TContext> : IGameEngine
     public TContext Context { get; private set; }
 
     /// <summary>
+    ///     The single guaranteed static fallback sentence for the engine error safety net (issue #271).
+    ///     This is the one acceptable canned string: it is returned only when the AI narrator itself
+    ///     cannot be reached (generation disabled, or it threw while producing the "oops" narration), so
+    ///     that turn processing still returns a graceful, in-character 200 body instead of a hard failure.
+    /// </summary>
+    public const string EngineErrorFallbackMessage =
+        "Something goes wrong, and for a moment the world seems to flicker. Perhaps try that a different way.";
+
+    /// <summary>
     ///     The flat list of held item names, used by the web clients to render the player's hands.
     ///     Derived live from <see cref="Context" />.Items rather than stored, so the projection can
     ///     never drift from the authoritative game state. A stored copy previously went stale on the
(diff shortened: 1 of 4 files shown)


arsindelve:
Thanks for the thorough review — went through each point:

- **Item 2 (broad `catch` also catching cancellation):** deliberately kept broad, and acting on the suggestion would actually be a regression. HttpClient timeouts (e.g. the OpenAI generation call) surface as `TaskCanceledException` (a subclass of `OperationCanceledException`), and a timeout is precisely the kind of deep failure this net must degrade gracefully. `GetResponse` has no `CancellationToken`, so there's no reliable way to tell a genuine client-cancel apart from a timeout — filtering cancellation out would let timeouts regress straight back into the HTTP 500s the net exists to prevent. I've added a comment at the catch site (78f0022) so this isn't "fixed" later.
- **Item 1 (double `TurnLog` write):** intentional. The `ENGINE ERROR (...)` row is the diagnostic record — it keeps failed turns visible in the Turns stream (aligned with the issue's "log loudly / stay discoverable" intent), while the `PostProcessing` row is the player-facing turn. They serve different purposes; happy to revisit if a downstream turn-counter cares.
- **Item 3 (multi-sentence partial output lost):** accepted scope trade-off — exactly the "safety net, not root-cause fix" boundary called out in the PR. The state changes from earlier sentences still happen; only the per-sentence confirmation text is collapsed into the single "oops" for this already-rare crash case.
- **Item 4:** agreed — it's inside the inner best-effort try, so worst case we just lose the location/score/moves on that one log line.

No behavior change in the follow-up commit — comment only.

---

Issue #222, opened 2026-06-21
Planetfall: Floyd lower-elevator-card reveal uses flat odds, not the original day/move escalation


arsindelve:
Floyd's reveal of his own lower-elevator card fires on flat per-turn odds. The original ramps the chance up over the game's days (and half-days) and **guarantees** it after Day 3, while never revealing on Day 1.

## What the ZIL does (ground truth)

`FLOYD-REVEAL-CARD-F` (`planetfall-source/globals.zil:1440-1468`), called each time you successfully swipe an elevator card with Floyd watching (`globals.zil:1391,1400`). With Floyd present and `CARD-REVEALED` not yet set, the reveal chance is:

| Day | Time of day | Chance |
|---|---|---|
| 2 | first half (`INTERNAL-MOVES < 5000`) | 5% |
| 2 | second half (`≥ 5000`) | 10% |
| 3 | first half | 20% |
| 3 | second half | 40% |
| > 3 | — | always (100%) |
| 1 | — | never (no clause matches) |

On success it sets `CARD-REVEALED` + `FLOYD-SPOKE` and either reveals/holds out the card ("Those cards are really neat, huh? Floyd has one for himself—see?") or, if the card was already stolen, mimes searching and comes up empty.

## C# current state

`FloydInventoryManager.OfferLowerElevatorCard` (`Planetfall/Item/Kalamontee/Mech/FloydPart/FloydInventoryManager.cs`):
```csharp
if (!IsFloydInRoom(context) || !floyd.IsOn || !floyd.Items.Any() || !chooser.RollDiceSuccess(3))
    return null;
```
That's a flat **~33% per qualifying turn**, gated only on Floyd being present/on and still holding the card internally. No `Day` gate, so it can reveal on **Day 1** (too early) and it **never guarantees** the reveal after Day 3. `PlanetfallContext.Day` and `CurrentTime` (the chronometer) both exist, so the day/half-day inputs are available.

## Fix

Replace the flat roll with a day-keyed escalating chance, using `context.Day`:
- Day 1 → never.
- Day 2 → ~5% (first half) / ~10% (second half).
- Day 3 → ~20% / ~40%.
- Day > 3 → always.

[...]

PR #287, opened 2026-06-23
Add Floyd SHOW reactions: printout/computer gate, cards, LLM default (#203)

### LINKED ISSUE #203: Floyd doesn't react to SHOW (printout/card puzzle reactions missing)
Showing objects to Floyd drives several reactions in the original — two real puzzle hooks plus flavor. The C# Floyd only handles GIVE, so `show <x> to floyd` does nothing at all.

> **Folds in #280** (the `show printout to floyd` → computer-concern gate). That reaction is the highest-value branch here and is **implementable today** — see "Puzzle hook 1" below. #280 is closed in favor of this issue.

> **Status: implemented** on `claude/hungry-wu-e92bf0`. One deliberate divergence from the ZIL ground truth below: the **default branch (#6) is an LLM-generated, in-character Floyd reaction**, not the fixed "Can you play any games with it?" line (which is retained only as a fallback). See Scope / Implementation.

`FLOYD-F`, the `SHOW` branch (`planetfall-source/compone.zil:2022-2047`) — `show <x> to floyd`. The branches are evaluated **top to bottom, first match wins**; order matters for the printout and lower-card guards:

| # | Condition (in order) | Response |
|---|---|---|
| 1 | `PRSO == PRINT-OUT` **and** `COMPUTER-FLAG` not yet set | runs `COMPUTER-ACTION` → Floyd's "computer is broken" concern (**sets `COMPUTER-FLAG`**; see Puzzle hook 1) |
| 2 | `PRSO == ROBOT-HOLE` | `FLOYD-THROUGH-HOLE` |
| 3 | `HERE == REC-AREA` **and** `PRSO == PSEUDO-OBJECT` (paddleball) | "Too intellectual for Floyd. Any paddleball sets around?" |
| 4 | `PRSO ∈ {ID-CARD, SHUTTLE-CARD, KITCHEN-CARD, UPPER-ELEVATOR-CARD}` | "Floyd scratches his head. \"Aren't those things usually blue?\"" |
| 5 | `PRSO == LOWER-ELEVATOR-CARD` **and** `CARD-REVEALED` not yet set | "\"I've got one just like that!\" says Floyd. He looks through several of his compartments, then glances at you suspiciously." (**sets `CARD-REVEALED`**; see Puzzle hook 2) |
| 6 | anything else | "Floyd looks over the \<x\>. \"Can you play any games with it?\" he asks." — *in the C# port this default is generated by the LLM in Floyd's voice (canned line kept as fallback); see Scope* |

Two consequences of the ordering and guards that the implementation must preserve:

- **Printout is first and one-shot.** Branch 1's guard is `<NOT ,COMPUTER-FLAG>`. Once the flag is set (by *either* trigger — see below), showing the printout again no longer fires `COMPUTER-ACTION`; the printout isn't a hole/pseudo/card, so it falls through to the **default** (branch 6, "Can you play any games with it?").
- **Lower card is one-shot too.** Branch 5's guard is `<NOT ,CARD-REVEALED>`. Once revealed (by SHOW *or* by Floyd's own reveal daemon), a second `show lower card to floyd` also falls through to the default.

### Puzzle hook 1 — printout → computer concern → bio-lab card foray (folds in #280)

This is the part #203 originally deferred by mistake. It is a genuine puzzle path, not flavor, and everything it needs already exists in the C# port.

`COMPUTER-ACTION` (`planetfall-source/comptwo.zil:1514-1524`):

```
<ROUTINE COMPUTER-ACTION ()
	 <SETG COMPUTER-FLAG T>
	 <SETG FLOYD-SPOKE T>
	 <TELL "Floyd examines the ">
	 <COND (<EQUAL? ,HERE ,COMPUTER-ROOM>
		<TELL "glowing light">)
	       (T
		<TELL "computer printout">)>
	 <TELL ". With a concerned frown, he says, \"Uh oh. Computer is
broken. A Doctor-person once told Floyd that Computer is the most important
part of the Project.\"" CR>>
```

`COMPUTER-ACTION` sets **`COMPUTER-FLAG`** and fires from **two** triggers:

1. **Floyd present in the Computer Room** — `KLUDGE`, `compone.zil:2291-2293` (`HERE == COMPUTER-ROOM`, flag not set).
2. **`show printout to floyd`** — the SHOW branch above, `compone.zil:2024-2026`. Works **anywhere**, no Computer-Room visit required.

Why `COMPUTER-FLAG` matters — the **bio-lab card foray** (the whole Floyd sacrifice sequence) is gated on it. When Floyd peers in the bio-lab window (`comptwo.zil:1747-1758`):

```
<COND (,COMPUTER-FLAG
       <SETG FLOYD-WAITING T>
       <TELL "...We'll need card there to fix computer. Hmmm... I know! Floyd will get card. ...">)
      (T
       <TELL "\"Ooo, look,\" he says. \"There's a miniaturization booth access card!\"">)>
```

With the flag set he offers to fetch the card (→ sacrifice). Without it, just "Ooo, look, a card!" — no offer. So a player who grabs the printout and **shows it to Floyd** (a valid original path) should unlock the bio-lab foray; today they can't.

**Wording subtlety (don't reuse the existing constant verbatim):** `COMPUTER-ACTION` branches its text on location — "glowing light" in the Computer Room, "computer printout" everywhere else. `FloydConstants.ComputerBroken` is hardcoded to the **"glowing light"** variant (correct for the Computer-Room trigger). The SHOW-printout path fires elsewhere, so it needs a **printout-variant** message ("Floyd examines the computer printout. With a concerned frown, he says, ...") — emitting the "glowing light" constant while the player holds a printout in another room would itself be a divergence. (In practice the printout SHOW only fires when not-yet-concerned, which means outside the Computer Room, so "computer printout" is always the right wording for this path.)

### Puzzle hook 2 — lower-elevator card → `CARD-REVEALED` (shared with #222)

Branch 5 sets `CARD-REVEALED`. The **same** flag gates Floyd's own card-reveal daemon (`FLOYD-REVEAL-CARD-F`, the subject of #222). Showing Floyd the card and Floyd spontaneously revealing his are two paths to the same one-time state — they must share one flag so neither double-fires.

## C# current state

- **SHOW is entirely unhandled.** `Floyd.RespondToMultiNounInteraction` (`Planetfall/Item/Kalamontee/Mech/FloydPart/Floyd.cs:300-312`) only runs the give-something engine, then falls to base. `show <x> to floyd` produces no Floyd reaction.
- **The printout exists** (correcting this issue's original claim that it "doesn't exist yet"): `Planetfall/Item/Feinstein/ComputerOutput.cs`, nouns `["output", "computer output", "printout"]`, with the correct disease-status-report text ending *"Malfunkshun in Sekshun 384! Sumuneeng reepaar roobot"*. It starts in the Computer Room (`ComputerRoom.Init()` → `StartWithItem<ComputerOutput>()`). So `show printout to floyd` already resolves to a real item — only the reaction is missing.
- **The gate flag exists and is already wired to its payoff.** `ComputerRoom.FloydHasExpressedConcern` (`Planetfall/Location/Lawanda/ComputerRoom.cs:12`) is the C# `COMPUTER-FLAG`. It is currently set in **one** place — `ComputerRoom.Act()` (`ComputerRoom.cs:61-65`, trigger #1, faithful). The bio-lock state machine already branches on it exactly as the ZIL branches on `COMPUTER-FLAG`: `BioLockEast.cs:77` passes it into `BioLockStateMachineManager.HandleTurnAction(...)`, where `BioLockStateMachineManager.cs:133-137` (`computerRoomFloydHasExpressedConcern && !FloydHasSaidNeedToGetCard` → `NeedToGetCard`) vs `:141` (the "look a mini card" no-offer path). **Trigger #2 (show printout) is the only missing setter.** This path is *not* blocked on #113 — that's the microbe fight, which happens later and is unrelated to this gate.
- **No `CARD-REVEALED` analog yet.** The lower-card reveal daemon `FloydInventoryManager.OfferLowerElevatorCard` (`FloydInventoryManager.cs:38-53`) gates only on Floyd present/on + still holding the card + a flat roll; there's no revealed flag (the subject of #222). One must be introduced and shared.

## Scope

**Implement now** — every item these need already exists:

- **Printout → computer concern (Puzzle hook 1).** `show printout to floyd` with Floyd present + on, when `ComputerRoom.FloydHasExpressedConcern` is not yet set → set it `true` and emit the **printout-variant** "computer is broken" message. This unlocks the bio-lab foray via the already-built state machine. *(This is the #280 fix.)*
- **"Aren't those usually blue?"** — for `IdCard`, `ShuttleAccessCard`, `KitchenAccessCard`, `UpperElevatorAccessCard`. **Enumerate these four explicitly.** Do **not** use `is AccessCard`: `IdCard` isn't even an `AccessCard` (it extends `ItemBase` directly — `Planetfall/Item/Feinstein/IdCard.cs`), while `TeleportationAccessCard`, `MiniaturizationAccessCard`, and `LowerElevatorAccessCard` *are* `AccessCard`-derived but must **not** get the blue line. A type check would both miss the id card and wrongly capture the teleport/mini/lower cards.
- **Lower-elevator card recognition (Puzzle hook 2).** `LowerElevatorAccessCard`, first time (revealed flag not set) → "I've got one just like that!" and **mark revealed**.
- **Default** — anything else shown to Floyd → an **LLM-generated, in-character reaction** (Floyd curiously looks the object over in his childlike way), *not* the fixed line. Since `RespondToMultiNounInteraction` has no `IGenerationClient`, the reaction is queued via the existing `CommentOnAction` path (rendered by Floyd's `Act()` that turn); the canned "Floyd looks over the … Can you play any games with it?" line is retained only as a fallback for when a comment can't be queued (Floyd already reacted to that object, already spoke this turn, or isn't a turn actor), so the player never gets a blank response.

**Defer (no corresponding objects in the port):**

- **robot hole → `FLOYD-THROUGH-HOLE`** and **rec-area paddleball pseudo → "Too intellectual…"** — no `RobotHole`/paddleball objects exist (only endgame narrative text references a paddleball set). Minor; revisit if those objects are ever modeled.

## Two shared-flag coordination points

1. **`FloydHasExpressedConcern` (= `COMPUTER-FLAG`).** The printout SHOW must set the **existing** `ComputerRoom.FloydHasExpressedConcern` (reached via `Repository.GetLocation<ComputerRoom>()`), **not** a new flag — so the Computer-Room visit and the show-printout path are the two interchangeable triggers the ZIL intends, and either one unlocks the bio-lab foray. Mirror `ComputerRoom.Act()`'s set-flag-and-return-message shape, but with the printout-variant string.
2. **"card revealed" flag (= `CARD-REVEALED`).** Coordinate with **#222**: introduce one shared revealed flag (on Floyd / `FloydInventoryManager`) that *both* the lower-card SHOW handler here and the reveal daemon set/check, so showing the card stops the daemon from also revealing it (and vice-versa).

## Implementation

Add SHOW handling to `Floyd.RespondToMultiNounInteraction`, **before** the give engine: match the verb (`show`/`display`) with the second noun resolving to Floyd (and Floyd on/present), resolve the first noun to the shown item, then branch in the ZIL's order — printout-first (guarded on `!FloydHasExpressedConcern`), then the four blue cards, then the lower card (guarded on the shared revealed flag), then the default. Fall through to the give engine / base only when no SHOW branch matches.

The default branch generates Floyd's reaction with the LLM: it queues an object-aware prompt (`FloydPrompts.ShownAnObject(noun)`) via `CommentOnAction` and returns an empty body so only Floyd's `Act()`-rendered line shows; if the comment can't be queued it returns the canned `FloydConstants.ShowDefaultFormat` line instead. SHOW also requires the shown item to be held (`syntax.zil:450` — `SHOW OBJECT (HAVE) …`), mirroring the give engine's "You don't have the X!" guard.

## TDD plan (per CLAUDE.md)

In `Planetfall.Tests/FloydTests.cs` (no SHOW coverage exists today), Floyd present (`IsOn`) and holding/with-access-to the relevant item; `Repository.Reset()`, mock `IRandomChooser`:

- `show printout to floyd` (Floyd present + on, `FloydHasExpressedConcern == false`) → printout-variant "computer is broken" message; assert `Repository.GetLocation<ComputerRoom>().FloydHasExpressedConcern == true` afterward, and that the bio-lock sequence then advances to `NeedToGetCard` (the #280 acceptance check). A second `show printout to floyd` → the default "play any games" line (flag already set).
- `show id card to floyd` (and one other of the four) → "usually blue". Add a guard test that a non-blue `AccessCard` (e.g. `TeleportationAccessCard` or `LowerElevatorAccessCard`) does **not** get the blue line.
- `show lower elevator card to floyd` (first time) → "just like that"; assert the shared revealed flag is now set, and that the #222 daemon won't re-reveal.
- `show <ordinary object> to floyd` → "Can you play any games with it?".

Each fails today (no SHOW reaction) and passes after.

## Acceptance criteria

- [x] `show printout to floyd` (Floyd present + on, not yet concerned) sets `ComputerRoom.FloydHasExpressedConcern` and emits the **printout-variant** concern message; the bio-lab foray then unlocks via the existing `BioLockStateMachineManager` (was #280)
- [x] Showing the printout again after the flag is set → the default "play any games" line (one-shot guard honored)
- [x] The four "blue" cards (`IdCard`, `ShuttleAccessCard`, `KitchenAccessCard`, `UpperElevatorAccessCard`) → the blue line, enumerated explicitly; teleport/mini/lower cards and the id card are handled correctly (no `is AccessCard`)
- [x] Lower-elevator card (first time) → recognition + revealed flag set; that flag is **shared** with the #222 reveal daemon (no double reveal)
- [x] Everything else shown → an LLM-generated in-character reaction (via `CommentOnAction`), with the canned line retained only as a no-blank fallback (e.g. repeat-show of the same object)
- [x] robot-hole / rec-area paddleball reactions explicitly deferred (no objects exist), noted not silently dropped
- [x] Deterministic tests added, red-before/green-after; `dotnet test Planetfall.Tests` green

## Affected files

- `Planetfall/Item/Kalamontee/Mech/FloydPart/Floyd.cs` — SHOW handling in `RespondToMultiNounInteraction` (+ shared `HasRevealedLowerElevatorCard` flag)
- `Planetfall/Item/Kalamontee/Mech/FloydPart/FloydConstants.cs` — printout-variant concern string + card / default strings
- `Planetfall/Item/Kalamontee/Mech/FloydPart/FloydPrompts.cs` — `ShownAnObject(noun)` LLM prompt for the default reaction
- `Planetfall/Item/Kalamontee/Mech/FloydPart/FloydInventoryManager.cs` — shared "card revealed" flag (coordinate with #222)
- `Model/Verbs.cs` — `ShowVerbs` (`show`/`display`)
- `Planetfall.Tests/FloydTests.cs` — proving tests
- `UnitTests/IntentMappings/base-mappings.json` — test-parser mappings for `show … to floyd`

---

_Verified against `planetfall-source/compone.zil:2022-2047` (`FLOYD-F` SHOW branch) and `:2291-2293` (Computer-Room trigger); `planetfall-source/comptwo.zil:1514-1524` (`COMPUTER-ACTION`) and `:1747-1758` (the `COMPUTER-FLAG` bio-lab gate). Folds in #280 (AB-021, found during AdventureBreaker bio-lock testing); both found via the ZIL behavior audit. Coordinates with #222 (shared `CARD-REVEALED` flag) and is independent of #113 (the microbe fight, downstream of this gate)._

### LINKED ISSUE #280: "show printout to floyd" should trigger Floyd's computer concern (bio-lab foray gate) — ZIL divergence
## Bug (ZIL-verified divergence)

Floyd's offer to fetch the miniaturization access card from the bio-lab — the whole sacrifice sequence — is gated on Floyd having expressed concern that "the computer is broken." In the original there are **two** ways to trigger that concern; the C# only implements one, so a legitimate original solution path is dead.

## ZIL ground truth

- **The gate** (`comptwo.zil:1751-1758`): when Floyd peers in the bio-lab window he only offers to get the card (*"We'll need card there to fix computer… I know! Floyd will get card."*) **if `COMPUTER-FLAG` is set**; otherwise just *"Ooo, look, a card!"* — no offer, no sacrifice.
- **What sets `COMPUTER-FLAG`** is `COMPUTER-ACTION` (`comptwo.zil:1514-1523`) — Floyd's *"Uh oh. Computer is broken. A Doctor-person once told Floyd that Computer is the most important part of the Project."* It fires from **two** triggers:
  1. Floyd present in the Computer Room — `compone.zil:2291-2293` (`HERE == COMPUTER-ROOM`).
  2. **`show printout to floyd`** — `compone.zil:2024-2026` (`SHOW` verb on `PRINT-OUT`), anywhere, no Computer-Room visit required.

## C# current state

- `FloydHasExpressedConcern` is set **only** in `Planetfall/Location/Lawanda/ComputerRoom.cs:64` (trigger #1). Faithful and correct.
- **Trigger #2 is missing.** The printout itself *is* modeled — `Planetfall/Item/Feinstein/ComputerOutput.cs` (nouns `["output","computer output","printout"]`, with the correct disease-status-report text ending *"Malfunkshun in Sekshun 384! Sumuneeng reepaar roobot"*) — but showing it to Floyd does nothing. `FloydHasExpressedConcern` is set nowhere else.

## Impact

A player who grabs the computer printout and **shows it to Floyd** (a valid original path) never unlocks the bio-lab card foray — so the central Floyd sacrifice becomes unreachable via that route. The Computer-Room-visit path still works, so it's not a hard block, but it diverges from the original and can leave a player stuck if they followed the show-printout approach.

## Suggested fix

Make `show printout to floyd` (the `ComputerOutput` item, with Floyd present + on) set `FloydHasExpressedConcern` and emit the `ComputerBroken` message, mirroring `ComputerRoom.Act()`. TDD: with Floyd present and the printout held, `show printout to floyd` → `FloydHasExpressedConcern == true` and the bio-lock sequence then progresses to `NeedToGetCard`.

Found by ZIL routine-level audit during AdventureBreaker bio-lock testing. AB-021.

arsindelve:
## What & why

Floyd never reacted to `show <x> to floyd` — only GIVE was handled, so the entire `FLOYD-F` SHOW branch (`compone.zil:2022-2047`) was missing. This ports it, in the ZIL's evaluation order. Closes #203, and folds in / closes #280 and #222.

## Reactions

| Shown to Floyd | Reaction |
|---|---|
| **Printout** (first time) | Sets `ComputerRoom.FloydHasExpressedConcern` (the `COMPUTER-FLAG` analog) and emits a "computer is broken" message — **unlocking the bio-lab card foray**. This is the #280 fix: showing the printout is the second, previously-missing trigger for that gate (the Computer-Room visit was the only one). |
| **The four "blue" cards** (id / shuttle / kitchen / upper-elevator) | "Aren't those usually blue?" |
| **Lower-elevator card** (first time) | "I've got one just like that!" + marks it revealed |
| **Anything else** | An **LLM-generated, in-character reaction** (see below) |

## Also folds in #222 — lower-card reveal escalation

The same lower-card daemon (`FloydInventoryManager.OfferLowerElevatorCard`) used a flat ~33% roll every qualifying turn. Based on `FLOYD-REVEAL-CARD-F` (`globals.zil:1440-1455`), the chance now escalates by day and is guaranteed after Day 3 — **Day 1 = 5%, Day 2 = 10%, Day 3 = 30%, Day > 3 = 100%** (via `RollDice(100)`).

- **Deliberate divergence:** the ZIL gives Day 1 a strict **0%**. Per a decision on this PR, we keep a small **non-zero** Day-1 chance so the card stays obtainable on Day 1 — the existing Day-1 walkthroughs (and early-game flow) get the card from Floyd before any sleep, which a hard 0% would break. The day-escalation and post-Day-3 guarantee (the substance of #222) are preserved.
- The original's `INTERNAL-MOVES`-based mid-day split has no clean C# analog (the chronometer resets to a morning base each day with no fixed day length), so each day uses one representative chance.
- The shared `HasRevealedLowerElevatorCard` flag (the #203 coordination) is unchanged.

## Notes for reviewers

- **`COMPUTER-ACTION` wording varies by location** (`comptwo.zil:1514-1524`): "glowing light" in the Computer Room vs "computer printout" elsewhere. The show-printout path fires elsewhere, so it uses a dedicated printout-variant string rather than reusing the existing "glowing light" `ComputerBroken` constant (which would have been a divergence).
- **Blue cards are enumerated explicitly**, *not* `is AccessCard` — `IdCard` isn't an `AccessCard` at all, while the teleport/mini/lower cards *are* but must not get the blue line.
- **Two shared one-shot flags**:
  - `ComputerRoom.FloydHasExpressedConcern` is now set by *both* the Computer-Room visit and the show-printout path (the bio-lock state machine already consumes it).
  - New `Floyd.HasRevealedLowerElevatorCard` (the `CARD-REVEALED` analog) is set/checked by *both* the lower-card SHOW and the spontaneous reveal daemon, so they never double-reveal (coordinates with #222).
- **LLM default (deliberate divergence from the canned ZIL line).** `RespondToMultiNounInteraction` has no `IGenerationClient`, so the default reaction is queued via the existing `CommentOnAction` path (rendered by Floyd's `Act()` that turn) using an object-aware prompt. The canned "Can you play any games with it?" line is retained **only as a fallback** for when a comment can't be queued (Floyd already reacted to that object, already spoke this turn, or isn't a turn actor) — so the player never gets a blank response.
- **SHOW requires the item to be held** (`syntax.zil:450` — `SHOW OBJECT (HAVE) …`), mirroring the give engine's possession guard.

## Tests (TDD per CLAUDE.md)

Deterministic tests in `Planetfall.Tests/FloydTests.cs`, confirmed red-before / green-after, covering each SHOW branch, both one-shot guards, the LLM default + its repeat-show fallback, the daemon-coordination flag (both directions), and the #222 day-escalation (Day-1 gettable, escalation across days, post-Day-3 guarantee — verified red against the old flat roll). The two Day-1 walkthroughs and the `DoesFloydOfferCard_*` integration tests were updated to force the reveal via the new `RollDice(100)` mechanism.

- **Planetfall.Tests**: 2320/2320
- **UnitTests**: 891/891 (1 pre-existing skip)
- **ZorkOne.Tests**: 1251/1251

`Model/Verbs.cs` (`ShowVerbs`) and `UnitTests/IntentMappings/base-mappings.json` are shared across games; the latter two suites are run as regression guards.

### DIFF
--- a/Model/Verbs.cs
+++ b/Model/Verbs.cs
@@ -39,6 +39,13 @@ public static class Verbs
     public static readonly string[] OpenVerbs = ["open", "lift", "pry", "unbar", "raise", "uncover", "unlatch", "unseal"];
     public static readonly string[] GiveVerbs =
         ["give", "offer", "transfer", "present", "provide", "hand", "donate", "deliver", "pass", "feed"];
+
+    /// <summary>
+    ///     The "show it to someone" family. Distinct from <see cref="GiveVerbs" /> — showing keeps the
+    ///     object in your hand (the ZIL syntax is <c>SHOW OBJECT (HAVE) TO OBJECT</c>), whereas giving
+    ///     transfers it. Deliberately excludes "present" (a give synonym) so the two verbs don't collide.
+    /// </summary>
+    public static readonly string[] ShowVerbs = ["show", "display"];
     public static readonly string[] ThrowVerbs =
         ["throw", "toss", "launch", "hurl", "fling", "heave", "chuck", "lob", "pitch", "cast", "sling"];
 
--- a/Planetfall.Tests/FloydTests.cs
+++ b/Planetfall.Tests/FloydTests.cs
@@ -1914,6 +1914,20 @@ public class FloydTests : EngineTestsBase
     }
 
     [Test]
+    public void ShownAnObjectPrompt_WithBracesInNoun_SurvivesDownstreamStringFormat()
+    {
+        // GenerateCompanionSpeech runs the prompt through string.Format; a shown noun with literal
+        // braces must not throw there (would be a FormatException at runtime) and must survive intact.
+        var prompt = FloydPrompts.ShownAnObject("gizmo {with} braces");
+
+        Action format = () => string.Format(prompt, "Room Name", "room description", "last outputs");
+
+        format.Should().NotThrow();
+        string.Format(prompt, "Room Name", "room description", "last outputs")
+            .Should().Contain("gizmo {with} braces");
+    }
+
+    [Test]
     public void Show_LowerCard_RevealedFlag_StopsTheDaemonReReveal()
     {
         // Coordination with #222: once the lower card is revealed (here, as if shown to Floyd), the
--- a/Planetfall.Tests/FloydTests.cs
+++ b/Planetfall.Tests/FloydTests.cs
@@ -1,6 +1,7 @@
 using System.Reflection;
 using FluentAssertions;
 using Model.AIGeneration.Requests;
+using Model.Intent;
 using Model.Interface;
 using Model.Item;
 using Moq;
@@ -1789,20 +1790,26 @@ public class FloydTests : EngineTestsBase
     }
 
     [Test]
-    public async Task Show_Printout_ToFloyd_WhenAlreadyConcerned_FallsToDefault()
+    public async Task Show_Printout_ToFloyd_WhenAlreadyConcerned_FallsToLlmDefault()
     {
+        // Already concerned (e.g. Floyd already visited the Computer Room): the one-shot guard
+        // (<NOT ,COMPUTER-FLAG>) means showing the printout again no longer re-fires the concern.
+        // In real play Floyd is a registered turn actor, so the default goes through the LLM path.
         var target = GetTarget();
         StartHere<RobotShop>();
-        // Already concerned (e.g. Floyd already visited the Computer Room): the one-shot guard
-        // (<NOT ,COMPUTER-FLAG>) means showing the printout again hits the default branch. Set the flag
-        // before Take<ComputerOutput>() — instantiating the Computer Room homes the printout there.
+        // Set the flag before Take<ComputerOutput>() — instantiating the Computer Room homes the printout there.
         GetLocation<ComputerRoom>().FloydHasExpressedConcern = true;
         Take<ComputerOutput>();
-        GetItem<Floyd>().IsOn = true;
+        var floyd = GetItem<Floyd>();
+        floyd.IsOn = true;
+        target.Context.RegisterActor(floyd);
+        Mock.Get(target.GenerationClient)
+            .Setup(c => c.GenerateCompanionSpeech(It.IsAny<CompanionRequest>()))
+            .ReturnsAsync("Floyd looks it over. \"What're all the squiggles?\"");
 
         var response = await target.GetResponse("show printout to floyd");
 
-        response.Should().Contain("Can you play any games with it");
+        response.Should().Contain("What're all the squiggles"); // LLM default, not the concern line
         response.Should().NotContain("Computer is broken");
     }
 
@@ -1836,6 +1843,72 @@ public class FloydTests : EngineTestsBase
     }
 
     [Test]
+    public async Task Show_ShuttleCard_ToFloyd_AsksIfTheyAreUsuallyBlue()
+    {
+        var target = GetTarget();
+        StartHere<RobotShop>();
+        Take<ShuttleAccessCard>();
+        GetItem<Floyd>().IsOn = true;
+
+        var response = await target.GetResponse("show shuttle card to floyd");
+
+        response.Should().Contain("usually blue");
+    }
+
+    [Test]
+    public async Task Show_UpperElevatorCard_ToFloyd_AsksIfTheyAreUsuallyBlue()
+    {
+        var target = GetTarget();
+        StartHere<RobotShop>();
+        Take<UpperElevatorAccessCard>();
+        GetItem<Floyd>().IsOn = true;
+
+        var response = await target.GetResponse("show upper elevator card to floyd");
+
+        response.Should().Contain("usually blue");
+    }
+
+    [Test]
+    public async Task Show_FloydTheCard_ReverseNounOrder_StillReacts()
+    {
+        // RespondToShow accepts Floyd as either noun ("show x to floyd" / "show floyd the x"). The forward
+        // order is covered by the other tests; this locks in the reverse branch (Floyd as NounOne).
+        var target = GetTarget();
+        StartHere<RobotShop>();
+        Take<IdCard>();
+        var floyd = GetItem<Floyd>();
+        floyd.IsOn = true;
+
+        var intent = new MultiNounIntent
+        {
+            Verb = "show",
+            NounOne = "floyd",
+            NounTwo = "id card",
+            Preposition = "",
+            OriginalInput = "show floyd the id card"
+        };
+        var result = await floyd.RespondToMultiNounInteraction(intent, target.Context);
+
+        result.Should().NotBeNull();
+        result!.InteractionMessage.Should().Contain("usually blue");
+    }
+
+    [Test]
+    public async Task Show_ObjectNotHeld_ToFloyd_SaysYouDontHaveIt()
+    {
+        // SHOW requires the item be held (syntax.zil:450 SHOW OBJECT (HAVE) ...). With the object only on
+        // the ground (resolvable but not in inventory), Floyd reports you don't have it — mirroring GIVE.
+        var target = GetTarget();
+        var robotShop = StartHere<RobotShop>();
+        robotShop.ItemPlacedHere(GetItem<Diary>()); // in the room, not the player's inventory
+        GetItem<Floyd>().IsOn = true;
+
+        var response = await target.GetResponse("show diary to floyd");
+
+        response.Should().Contain("You don't have the");
+    }
+
+    [Test]
     public async Task Show_LowerElevatorCard_ToFloyd_RecognizesItAndMarksRevealed()
     {
         var target = GetTarget();
@@ -1854,17 +1927,24 @@ public class FloydTests : EngineTestsBase
     }
 
     [Test]
-    public async Task Show_LowerElevatorCard_ToFloyd_WhenAlreadyRevealed_FallsToDefault()
+    public async Task Show_LowerElevatorCard_ToFloyd_WhenAlreadyRevealed_FallsToLlmDefault()
     {
+        // Already revealed: the one-shot guard means a second show no longer gets the recognition line.
+        // In real play Floyd is a registered turn actor, so the default goes through the LLM path.
         var target = GetTarget();
         StartHere<RobotShop>();
         Take<LowerElevatorAccessCard>();
-        GetItem<Floyd>().IsOn = true;
-        GetItem<Floyd>().HasRevealedLowerElevatorCard = true;
+        var floyd = GetItem<Floyd>();
+        floyd.IsOn = true;
+        floyd.HasRevealedLowerElevatorCard = true;
+        target.Context.RegisterActor(floyd);
+        Mock.Get(target.GenerationClient)
+            .Setup(c => c.GenerateCompanionSpeech(It.IsAny<CompanionRequest>()))
+            .ReturnsAsync("Floyd peers at it. \"Ooo, shiny card!\"");
 
         var response = await target.GetResponse("show lower elevator card to floyd");
 
-        response.Should().Contain("Can you play any games with it");
+        response.Should().Contain("Ooo, shiny card"); // LLM default, not the recognition line
         response.Should().NotContain("just like that");
     }
 
--- a/Planetfall.Tests/FloydTests.cs
+++ b/Planetfall.Tests/FloydTests.cs
@@ -1763,4 +1763,195 @@ public class FloydTests : EngineTestsBase
     }
 
     #endregion
+
+    #region Show (issue #203 — Floyd's SHOW reactions, ZIL FLOYD-F compone.zil:2022-2047)
+
+    [Test]
+    public async Task Show_Printout_ToFloyd_TriggersComputerConcern()
+    {
+        var target = GetTarget();
+        StartHere<RobotShop>();
+        // Instantiate the Computer Room first: its Init() homes the printout there, so we must do this
+        // before Take<ComputerOutput>() or the Take would be undone (printout re-homed out of inventory).
+        GetLocation<ComputerRoom>().FloydHasExpressedConcern = false;
+        Take<ComputerOutput>();
+        GetItem<Floyd>().IsOn = true;
+
+        var response = await target.GetResponse("show printout to floyd");
+
+        // The "computer is broken" concern, in the printout-variant wording — NOT the Computer-Room
+        // "glowing light" wording (COMPUTER-ACTION branches its text on location, comptwo.zil:1514-1524).
+        response.Should().Contain("Computer is broken");
+        response.Should().Contain("computer printout");
+        response.Should().NotContain("glowing light");
+        // Sets the SAME flag the Computer-Room visit sets — this is what gates the bio-lab card foray.
+        GetLocation<ComputerRoom>().FloydHasExpressedConcern.Should().BeTrue();
+    }
+
+    [Test]
+    public async Task Show_Printout_ToFloyd_WhenAlreadyConcerned_FallsToDefault()
+    {
+        var target = GetTarget();
+        StartHere<RobotShop>();
+        // Already concerned (e.g. Floyd already visited the Computer Room): the one-shot guard
+        // (<NOT ,COMPUTER-FLAG>) means showing the printout again hits the default branch. Set the flag
+        // before Take<ComputerOutput>() — instantiating the Computer Room homes the printout there.
+        GetLocation<ComputerRoom>().FloydHasExpressedConcern = true;
+        Take<ComputerOutput>();
+        GetItem<Floyd>().IsOn = true;
+
+        var response = await target.GetResponse("show printout to floyd");
+
+        response.Should().Contain("Can you play any games with it");
+        response.Should().NotContain("Computer is broken");
+    }
+
+    [Test]
+    public async Task Show_IdCard_ToFloyd_AsksIfTheyAreUsuallyBlue()
+    {
+        // IdCard extends ItemBase directly (NOT AccessCard) — proves the four "blue" cards must be
+        // enumerated explicitly rather than matched via `is AccessCard`.
+        var target = GetTarget();
+        StartHere<RobotShop>();
+        Take<IdCard>();
+        GetItem<Floyd>().IsOn = true;
+
+        var response = await target.GetResponse("show id card to floyd");
+
+        response.Should().Contain("usually blue");
+    }
+
+    [Test]
+    public async Task Show_KitchenCard_ToFloyd_AsksIfTheyAreUsuallyBlue()
+    {
+        // KitchenAccessCard IS an AccessCard — proves the enumeration spans both type hierarchies.
+        var target = GetTarget();
+        StartHere<RobotShop>();
+        Take<KitchenAccessCard>();
+        GetItem<Floyd>().IsOn = true;
+
+        var response = await target.GetResponse("show kitchen card to floyd");
+
+        response.Should().Contain("usually blue");
+    }
+
+    [Test]
+    public async Task Show_LowerElevatorCard_ToFloyd_RecognizesItAndMarksRevealed()
+    {
+        var target = GetTarget();
+        StartHere<RobotShop>();
+        Take<LowerElevatorAccessCard>();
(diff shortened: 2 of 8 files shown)


arsindelve:
Thanks for the thorough review! Addressed the two minor items in `4eed709`:

- **Brace-escaping fragility in `ShownAnObject`** — the prompt now escapes `{`/`}` in the shown item's noun before it reaches `GenerateCompanionSpeech`'s `string.Format` pass, so a future noun containing braces can't throw a `FormatException`. Added a regression test (`ShownAnObjectPrompt_WithBracesInNoun_SurvivesDownstreamStringFormat`).
- **`Repository.GetLocation<ComputerRoom>()` lookup** — moved inside the `shown is ComputerOutput` branch so it's only performed for the printout, not on every show. Behavior unchanged.

No change for the affirmations / informational notes: the SHOW-before-GIVE ordering is intentionally defensive (you're right it isn't load-bearing today, but it documents intent and guards against a future shared synonym), and the branch logic / flag sharing were confirmed correct.

`Planetfall.Tests` 2311/2311 green after the change.

arsindelve:
Thanks again! Rounded out the test coverage in `3dadbc2` — all four suggestions addressed:

- **Untested blue cards** — added `Show_ShuttleCard_ToFloyd_AsksIfTheyAreUsuallyBlue` and `Show_UpperElevatorCard_ToFloyd_AsksIfTheyAreUsuallyBlue`, so all four explicitly-enumerated cards are now exercised.
- **Reverse noun order** — added `Show_FloydTheCard_ReverseNounOrder_StillReacts` (constructs a `MultiNounIntent` with Floyd as `NounOne`), locking in the symmetric branch.
- **Missing-item branch** — added `Show_ObjectNotHeld_ToFloyd_SaysYouDontHaveIt` (object on the ground, not in inventory → "You don't have the …").
- **Already-revealed/concerned default going through the LLM path in production** — good catch. Made both `WhenAlready…` tests production-faithful: they now `RegisterActor(floyd)` (as `FloydPowerManager.Activate` does in real play) and mock `GenerateCompanionSpeech`, so they assert the LLM default fires (and the one-shot message does *not*), instead of the canned fallback. The canned fallback remains covered by the diary repeat-show test.

`Planetfall.Tests` 2315/2315, `UnitTests` 891/891 (1 pre-existing skip), `ZorkOne.Tests` 1251/1251 — all green.

arsindelve:
Thanks for the careful read! Here's how I handled the three observations:

- **Indirect success signal for the queued comment** → fixed in `e7ddfb9`. `CommentOnAction` now returns a `bool` (true when queued, false on its no-op paths), and `RespondToShow` branches on that return instead of inferring success from `PendingFloydActionCommentPrompt == reactionPrompt`. The seven existing callers ignore the return, so it's source-compatible. Planetfall.Tests 2315/2315 still green.

- **Duplication with `GiveSomethingToSomeoneDecisionEngine<T>`** → not changed, agreeing with your read. Extracting a `ShowSomethingToSomeoneDecisionEngine<T>` now would be premature abstraction for a single Floyd-only consumer; the right moment is when a second NPC actually needs SHOW reactions, at which point the give engine is the obvious template to mirror.

- **Trailing-space nit on `ShowDefaultFormat` / `CardsUsuallyBlue`** → left as-is. The response combiner `TrimEnd`s each part before joining, so it has no output effect, and (as you noted) the file's convention is already mixed — `ComputerBroken`, the sibling these mirror, has no trailing space either. Changing it would be arbitrary churn.

arsindelve:
Thanks for the thorough pass — no code changes this round, since all three notes are explicitly non-actionable for this PR:

- **`ComputerRoom.Init()` item-homing quirk** — agreed it's a pre-existing engine behavior, out of scope here. It's real (the `Show_Printout_…` setup comment documents it), so a follow-up issue makes sense; happy to file one if useful.
- **`ComputerBrokenFromPrintout` vs `ComputerBroken`** — leaving as two strings: only the two location-dependent variants exist (`comptwo.zil:1514-1524`), and a shared format string is the right move only once a third variant appears, per your note.
- **ZIL citations** — confirmed: I verified the cited `compone.zil` / `comptwo.zil` line ranges against the local `planetfall-source/` checkout while implementing; they're accurate (the reference checkout just isn't present in your sandbox).

Confirming green on the merge commit matches my local runs: `Planetfall.Tests` 2315/2315. Appreciate the review!

arsindelve:
Thanks! Both non-blocking items handled:

- **Off / not-in-room SHOW coverage** — added `Show_ToFloyd_WhenOff_ProducesNoShowReaction` (`6fb1c7d`). Confirmed the `!IsOn || HasDied` guard at `Floyd.cs:314` is upstream of `RespondToShow` (line 320), so the SHOW branches can't fire while Floyd is off; the new test locks that in. The not-in-room case is handled by the multi-noun engine's scoping — it only invokes `RespondToMultiNounInteraction` on items actually in the room/inventory, so an absent Floyd's handler is never reached.
- **Brace-in-format-string risk in other prompt builders** — checked; no risk. `ShownAnObject` is the *only* prompt that interpolates a runtime value into the prompt text. Every other Floyd prompt (`LaserPickedUp`, `MagnetRetrievesKey`, `PadlockUnlocked`, …) is a parameterless static string, and the only braces in `FloydPrompts.cs` are the intentional `{0}/{1}/{2}` placeholders in the AI prompt templates that `GenerateCompanionSpeech` is *meant* to format. So no follow-up needed there.

The percent-table simplification is documented with the `#222` cross-reference as you noted. `Planetfall.Tests` 2321/2321 green.

PR #295, opened 2026-06-23
Rebuild Floyd's idle-comment voice (canon-light) + random-seed variety

arsindelve:
## Why

Floyd's AI **idle comments** (the `PerformRandomAction` buckets) had drifted off his canonical voice. Playing in prod, his random lines came out as **fluent, wistful, adult prose** — *"Floyd wonders if his circuit board might get outdated and then nobody will need him"*, *"That panel stays dim because nobody resets it anymore"* — instead of the simple, eager, third-person child-speak that *is* Floyd.

Two root causes, both in `FloydPrompts.cs`:
1. The `SystemPrompt` globally instructed a *"slightly melancholic"* tone and to *"focus on how things… break, or wear down over time"* — so even the happy buckets came out moody.
2. Every dialog bucket said *"Floyd uses normal English sentence structure"* and its **examples were fluent adult prose**, training the model away from his voice. (Proof it was the prompt, not the model: the `LeavingToExplore`/`ReturningFromExploring` prompts in the same file — which model the broken voice — produce perfect Floyd: *"Floyd going exploring. See you later."*)

On top of that, output **clustered hard** (with identical room+context the model ruts — all rust, all balancing), occasionally **gave objects feelings**, **picked up room items**, and **over-used tag questions**.

This was developed empirically against `FloydConstants.cs` as the oracle (the canonical 19 + the scripted beats) using live `gpt-4o`/`gpt-4.1` generations, iterating prompts until generated output read like canon across 300+ samples.

## What changed

**`FloydPrompts.cs`**
- **New `SystemPrompt`** targeting the real canon register: simple, clean, eager, third-person, with a *light, occasional* quirk — explicitly **not** heavy pidgin ("Jar-Jar") and **not** adult-wistful. Uses the actual canon line as the exemplar (*"Uh oh. Computer is broken. A Doctor-person once told Floyd…"*).
- **Rewrote the 5 idle buckets**: childlike self-fret (not existential), childlike *"the people are gone"* melancholy (not adult elegy), terse eccentric actions. Hard guards added: **no anthropomorphizing**, **no touching/picking up room items**, **tag questions rare**, **one line only**.
- **Added `Seeds`** — 24 "what's on Floyd's mind right now" prompts per bucket.

**`Floyd.cs` (`PerformRandomAction`)**
- Each AI bucket now appends a **random seed** to its prompt, so identical room+context still varies. This mirrors how the canonical 19 already use `Chooser.Choose`.
- **Re-weighted** the buckets to lean on the canonical 19 (Floyd's eccentric heart, 35%) and keep **melancholy a thin ~8% slice**.

**`QuirkyCompanion.cs`**
- `GenerateCompanionSpeech` takes an optional `seed` and appends it to the prompt.

**`ChatGPTClient.cs` / `OpenAIClientBase.cs`**
- Companion speech runs on a **cheaper, equally-fast model** (`CompanionModelName`), scoped to the companion path only — **the narrator is untouched**. At ~1,130 input tokens vs ~25 output, this measured **~7× cheaper per call at equal latency/quality**.
- Curly quotes are normalized to straight quotes.
- ⚠️ **Confirm the companion model id (`gpt-5.4-mini`) is enabled on the account before deploying**, or change the one constant.

## Tests
`Planetfall.Tests/FloydTests.cs` (deterministic, mocked `IRandomChooser` + stubbed client):
- the chosen seed reaches the prompt the model sees;
- weighted bucket routing picks a seeded AI bucket and threads the seed through;
- every AI bucket has ≥20 seeds so a short run never repeats one.

Full `Planetfall.Tests` suite green (2306 passed); solution builds clean.

## Before → after (measured)
| | before | after |
|---|---|---|
| voice | fluent/wistful adult | canon-light child-speak |
| creativity (distinct openings / 100) | 81% w/ 5× ruts | 95% w/ flat 2× |
| anthropomorphizing / pick-ups / tag-tics | present | 0 across 100s of samples |
| companion cost / 1k lines | ~$2.43 (gpt-4o/4.1) | ~$0.34 |

https://claude.ai/code/session_01DsEMy2gYe2qexc95aJR4KR

---
_Generated by [Claude Code](https://claude.ai/code/session_01DsEMy2gYe2qexc95aJR4KR)_


mlane-sn (approved):

arsindelve:
Thanks for the thorough review — all good catches. Addressed in `7038ee0`:

**Blocking — model id + graceful degradation.** Two parts:
- The id `gpt-5.4-mini` **does resolve** — the whole voice/variety rework was developed against it on the live API (~0.8s/call, valid completions, ~7× cheaper than gpt-4o). It's a real model on the account, just an unusual-looking name. I kept the `NOTE: confirm … before deploying` comment as a guardrail.
- You're right that there was no graceful fallback. `GenerateCompanionSpeech` now wraps the call so a failing/misconfigured companion model **degrades to the narrator model** instead of throwing (`catch … when (CompanionClient is not null && Client is not null)`), in keeping with the codebase's degradation principle. So even if the id were ever disabled, Floyd keeps talking on `gpt-4o` rather than crashing the turn.

**Stale `RollDice(6)` stubs.** Fixed all three (`Act_PerformsNormally_WhenNoPromptPending`, `Act_ReturnsEmpty_WhenSkipActingFlagSet`, `Act_WorksNormally_AfterFlagReset`) — they now stub `RollDice(100)` with explicit values (`1` → the `<= 35` canonical branch they intend; `50` for the skip-flag test that never reaches the dice). No more relying on Moq's default-0 coincidence.

**Quote-normalization test gap.** Extracted the curly-quote logic to a pure `public static ChatGPTClient.NormalizeQuotes(string)` and covered it in `UnitTests/ChatGPTClientTests.cs` (curly double quotes, apostrophes, straight-quote passthrough). I left the `CompanionClient`/null-`ApiKey` branch untested for now since `ChatGPTClient`'s ctor requires a key (can't easily construct the no-key path in isolation) — happy to add an `InternalsVisibleTo` + seam if you'd like it covered.

**Minor:** restored the trailing newline in `FloydPrompts.cs`. Left the `ApiKey` exposure as-is (one subclass today) and the `.ToList()` hot-path call (existing pattern) per your notes.

Local: solution builds clean; Planetfall.Tests 2327/2327, UnitTests 934/935 (1 pre-existing skip).

---

arsindelve:
Thanks — noting this round found **no bugs/correctness issues**, so I'm not pushing another commit for the non-blocking suggestions (each push re-triggers this review, and I'd rather not churn). Quick dispositions so the loop is closed:

- **(4) Canon singing — intentional, not a contradiction.** The static `RandomActions` lines (`sings an ancient ballad`, `chants the death scene from Carmen`) are beloved original canon and stay. The `SystemPrompt`'s no-music rule deliberately governs only the **LLM-generated** idle buckets, where unprompted singing read as off. Both coexisting is by design.
- **(1) env-config the model id / (2) CloudWatch counter on the fallback / (3) test the fallback path** — all reasonable, all optional, none blocking. (3) needs a mockable `ChatClient` seam (the SDK type is constructed internally), so it's a small refactor rather than a quick add. Happy to do any of these if you want them — say the word and I'll pick them up; otherwise leaving them as maintainer's-call polish.
- **(5) `??=` race** — real but harmless (worst case constructs two cheap clients, last wins). Low priority.

The model id `gpt-5.4-mini` is verified working against the live API (per the earlier comment); the `catch … falls back to the narrator model` keeps companion lines working even if it's ever disabled. CI green, no blocking items — ready from my side unless you want any of the polish above.

---

PR #305, opened 2026-06-23
Fix #17: death-as-spirit mechanic (become a ghost at the gates of Hades)

### LINKED ISSUE #17: Dying after praying at the altar is a totally different experience
The C# port has no "death-as-spirit" mechanic at all: every death reincarnates you in the Forest. In the original, once you've **visited the altar (South Temple)**, dying turns you into a **ghost at the Entrance to Hades** — the "totally different experience" below — and you resurrect by walking back to the altar and praying.

## Expected behavior (original game), as reported

```
As you take your last breath, you feel relieved of your burdens. The feeling passes as you find
yourself before the gates of Hell, where the spirits jeer at you and deny you entry. Your senses are
disturbed. The objects in the dungeon appear indistinct, bleached of color, even unreal.

Entrance to Hades
There is a brass bell here.
There is a pair of candles here.

>i
You have no possessions.

>diagnose
You are dead.

>look
The room looks strange and unearthly and objects appear indistinct.

>wait
Might as well. You've got an eternity.

>score
You're dead! How can you think of your score?
```

## What the original game does (ground truth)

`JIGS-UP` — the death routine, `zork1/1actions.zil:4046`:

```zil
<COND (,DEAD                                  ; already a spirit -> permanent game over
       <TELL "It takes a talented person to be killed while already dead. ...">
       <FINISH>)>
...
<SCORE-UPD -10>                               ; death costs 10 points
<COND
 (<NOT <L? ,DEATHS 2>>                        ; this is the 3rd death -> permanent game over
  <TELL "You clearly are a suicidal maniac. ... Land of the Living Dead ...">
  <FINISH>)
 (T
  <SETG DEATHS <+ ,DEATHS 1>>
  <COND (<FSET? ,SOUTH-TEMPLE ,TOUCHBIT>      ; HAVE you visited the altar room?
         <TELL "As you take your last breath, you feel relieved of your burdens. ...
                before the gates of Hell ... objects appear indistinct, bleached of color, even unreal.">
         <SETG DEAD T>
         <SETG ALWAYS-LIT T>
         <PUTP ,WINNER ,P?ACTION DEAD-FUNCTION>   ; spirit verb handler
         <GOTO ,ENTRANCE-TO-HADES>)
        (T                                     ; never been to the altar -> ordinary reincarnation
         <TELL "Now, let's take a look here... you probably deserve another chance. ...">
         <GOTO ,FOREST-1>)>
  <RANDOMIZE-OBJECTS>                          ; scatter your inventory across the map
  <KILL-INTERRUPTS>)>
```

So the death outcome depends on three things:

1. **Already a spirit (`DEAD`)** → hard game over.
2. **3rd death (`DEATHS >= 2`)** → hard game over ("suicidal maniac… Land of the Living Dead"). You get exactly two reincarnations.
3. Otherwise, **if you've ever entered the altar room** (`SOUTH-TEMPLE`; the `ALTAR` object is `(IN SOUTH-TEMPLE)`, `zork1/1dungeon.zil:205`) → become a **spirit** at the Entrance to Hades. If not, the familiar Forest reincarnation.
4. Either way, `-10` points and `RANDOMIZE-OBJECTS` scatters your stuff (`zork1/1actions.zil:4101`: lamp → Living Room, coffin → Egyptian Room, treasures → random above-ground rooms, everything else → random above-ground rooms).

**Spirit behavior** — `DEAD-FUNCTION`, `zork1/1actions.zil:3113` — while `DEAD`, verbs are overridden:

| verb | response |
|---|---|
| wait | "Might as well. You've got an eternity." |
| score | "You're dead! How can you think of your score?" |
| inventory / drop / throw | "You have no possessions." |
| take / rub | "Your hand passes through its object." |
| diagnose | "You are dead." |
| look | "The room looks strange and unearthly[ and objects appear indistinct]." (+ dim light) |
| attack / mung / swing | "All such attacks are vain in your condition." |
| open/close/eat/drink/turn/burn/… | "Even such an action is beyond your capabilities." |
| turn on lamp | "You need no light to guide you." |
| anything else | "You can't even do that." |

**Resurrection** — `DEAD-FUNCTION`, `zork1/1actions.zil:3152`: while a spirit, go back to the altar and `PRAY`:

```zil
(<VERB? PRAY>
 <COND (<EQUAL? ,HERE ,SOUTH-TEMPLE>
        <SETG DEAD <>> <SETG ALWAYS-LIT <>> <PUTP ,WINNER ,P?ACTION 0>
        <TELL "From the distance the sound of a lone trumpet is heard. ... you find yourself
               rising as if from a long sleep, deep in the woods. ...">
        <GOTO ,FOREST-1>)
       (T <TELL "Your prayers are not heard.">)>)
```

While a spirit you are also always-lit ("Although there is no light, the room seems dimly illuminated").

## What the C# port does today (the bug)

`ZorkOne/Command/DeathProcessor.cs` — **every** death is the Forest reincarnation, unconditionally:

```csharp
zorkContext.DeathCounter++;
zorkContext.LightWoundCounter = 0;
var newLocation = Repository.GetLocation<ForestOne>();
context.CurrentLocation = newLocation;

// TODO: Scatter inventory.
// TODO: Die ?? number of times: Don't get another chance.
// https://github.com/historicalsource/zork1/blob/.../1actions.zil#L4046

var result = death + "\n\t*** You have died ***\n\n" +
             "Now, let's take a look here... Well, you probably deserve another chance. ...";
```

- No `DEAD`/spirit state exists on `ZorkIContext`. There is no `DEAD-FUNCTION` equivalent.
- No spirit death — visiting the altar then dying still just dumps you in the Forest (the TODOs even point at `JIGS-UP`).
- No death-count limit — you can die forever; the 3rd-death permanent game over is missing.
- No `RANDOMIZE-OBJECTS` — inventory is never scattered.
- `ZorkOne/Location/Altar.cs` handles `pray` only for the **living** case (teleport to `ForestOne`); there's no spirit-resurrection branch.
- The `Spirits` / `EntranceToHades` code is the separate **exorcism** puzzle (evil spirits bar the gate); it is unrelated to *becoming* a spirit.
- The trigger only needs "have you entered the altar room," which the engine already supports: every `LocationBase` exposes `VisitCount` (`GameEngine/Location/LocationBase.cs:167`; `VisitCount == 0` means never visited).

## How to fix

This is a feature, best staged:

1. **Spirit state:** add `bool IsDead` (spirit) to `ZorkIContext`.
2. **Altar-visited check:** use the existing per-room visit counter — `Repository.GetLocation<Altar>().VisitCount > 0` (every `LocationBase` tracks `VisitCount`). No new flag needed.
3. **Rework `DeathProcessor.Process`:**
   - if `IsDead` already → permanent game over ("It takes a talented person to be killed while already dead…").
   - `AddPoints(-10)`.
   - if `DeathCounter >= 2` (this is the 3rd death) → permanent game over ("You clearly are a suicidal maniac… Land of the Living Dead.").
   - else `DeathCounter++`; if `HasVisitedAltar` → spirit death: print the "relieved of your burdens… gates of Hell… indistinct" text, set `IsDead = true`, move to `EntranceToHades`, mark always-lit; else → current Forest reincarnation.
   - scatter inventory (lamp → `LivingRoom`, coffin → `EgyptianRoom`, treasures → random above-ground rooms, rest → random above-ground rooms). Use `IRandomChooser` so it's testable.
4. **Spirit verb behavior (`DEAD-FUNCTION`):** when `IsDead`, intercept commands at a high level (a global pre-processor / context hook) and emit the table above; block normal interaction.
5. **Resurrection:** when `IsDead` and at the `Altar`, `pray` → clear `IsDead` + always-lit, restore the lamp, print the trumpet/"rising as if from a long sleep" text, move to `ForestOne`. (Extend `Altar`'s pray handler with an `IsDead` branch; elsewhere `pray` → "Your prayers are not heard.")
6. **Light while dead:** treat the player as always-lit ("the room seems dimly illuminated").

## Tests (TDD — required per CLAUDE.md)

Deterministic (`Repository.Reset()`, mock `IRandomChooser` for the scatter):

- Visit `Altar`, then die → response contains "gates of Hell"/"indistinct", `CurrentLocation` is `EntranceToHades`, `IsDead == true`.
- Die **without** visiting the altar → Forest reincarnation (current behavior preserved).
- Third death → permanent game-over message; game ends.
- As a spirit: `wait`/`score`/`inventory`/`diagnose`/`look`/`take` each return the `DEAD-FUNCTION` message.
- As a spirit, `pray` at the `Altar` → reborn in `ForestOne`, `IsDead == false`.
- On death, inventory is scattered (e.g., lamp ends up in `LivingRoom`).

arsindelve:
## What & why

Closes #17. The C# port had no "death-as-spirit" mechanic — *every* death reincarnated the player in the forest. In the original (ZIL `JIGS-UP` / `DEAD-FUNCTION`), once you have visited the altar (South Temple), dying turns you into a **ghost at the Entrance to Hades**, and you resurrect by walking back to the altar and praying.

## Behavior now (faithful to `JIGS-UP`)

- **Killed while already a spirit** → permanent game over ("…killed while already dead…").
- **Third death** (after two reincarnations) → permanent game over ("suicidal maniac… Land of the Living Dead"). Still costs 10 points (matching `JIGS-UP`, where `SCORE-UPD -10` precedes the `FINISH`) but does **not** scatter. The *already-a-spirit* case skips even the point loss (its `FINISH` is above `SCORE-UPD`).
- **Otherwise:** lose **10 points**, the inventory **scatters**, and:
  - **visited the altar** → become a **spirit** at the Entrance to Hades.
  - **never visited** → the familiar **forest** reincarnation (unchanged).
- **As a spirit:** always-lit (no grue while wandering dark rooms), most verbs overridden (`wait`/`score`/`inventory`/`diagnose`/`look`/`take`/`attack`/…); **movement still works** so you can walk back.
- **Resurrection:** `pray` at the altar clears the spirit state and lifts you to the forest; `pray` elsewhere → "Your prayers are not heard."

Trigger uses the existing per-room `VisitCount` (no new flag). Scatter (`RANDOMIZE-OBJECTS`): lamp → Living Room, coffin → Egyptian Room, everything else → a random above-ground room via the mockable `IRandomChooser`.

## Implementation notes

- New generic `IContext.InterceptPlayerCommand` hook (default no-op) — lets a game-specific context fully handle a raw command before normal parsing while letting movement/resurrection fall through (`null`). `ItIsDarkHere` made `virtual`.
- `SpiritProcessor` is the `DEAD-FUNCTION` equivalent; `DeathProcessor` reworked with the branch logic + scatter.
- Trap documented in code: resolving a not-yet-visited scatter destination lazily runs that room's `Init`, which re-places its own starting items (e.g. the lamp) without clearing the inventory copy — so the scatter explicitly drops each item from the player first.
- A true "FINISH" (refuse all further input) isn't a concept this engine supports, so the permanent game-over cases return the terminal message rather than hard-locking input — consistent with how death already worked (relocate + message).

## Tests (TDD)

15 deterministic tests in `ZorkOne.Tests/Places/SpiritDeathTests.cs` cover every branch (spirit death, forest reincarnation, third death, -10 points, scatter, spirit verbs, resurrection, prayer-not-heard, always-lit, killed-while-dead).

No regressions: ZorkOne.Tests (1278), UnitTests (934), Planetfall.Tests (2343), EscapeRoom.Tests (7) — all pass; clean build, 0 warnings.


ZorkOne/ZorkIContext.cs
```diff
@@ -112,4 +138,41 @@ public override bool HaveRoomForItem(IItem item)
...
+    ///     back toward the altar (original ZIL: ALWAYS-LIT is set when DEAD). See issue #17.
+    /// </summary>
+    public override bool ItIsDarkHere => !IsDead && base.ItIsDarkHere;
+
+    /// <summary>
+    ///     The in-character "the game is over" line shown for every (non-system) command once the player
+    ///     has died permanently. The original game's FINISH prompt is not expressible here, so this is the
+    ///     stand-in: it ends the player's ability to act while leaving save/restore/quit available.
+    /// </summary>
+    public const string PermanentDeathMessage =
```
arsindelve, line 160: IS this correct per the ZIL?

ZorkOne/Command/SpiritProcessor.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        // Praying is how a spirit is resurrected — but only at the altar.
+        if (firstWord is "pray")
+            return Pray(context);
+
+        // Lighting a lamp: a spirit needs no light to guide it.
+        if (Matches(command, "turn on", "light", "activate") && MentionsLamp(command))
+            return "You need no light to guide you. ";
+
+        return firstWord switch
```
arsindelve: Are these verified against the ZIL?

ZorkOne/Command/SpiritProcessor.cs
```diff
@@ -0,0 +1,93 @@
...
+        {
```
arsindelve: Consider using VERBS.cs here

ZorkOne/ZorkIContext.cs
arsindelve: Verified — yes, this is faithful. I pulled `1actions.zil` from the pinned historicalsource commit:

- Becoming a spirit sets the global flag: `<SETG ALWAYS-LIT T>` (`1actions.zil:4086`).
- The `DEAD-FUNCTION` `LOOK` branch confirms the intent: "Although there is no light, the room seems dimly illuminated." (`:3157`).
- Resurrection clears it again: `<SETG ALWAYS-LIT <>>` (`:3159`).

So forcing `ItIsDarkHere => false` while `IsDead` is the correct port of `ALWAYS-LIT` (and it's what keeps a wandering spirit from being eaten by a grue). Added these citations to the doc comment in 1293046.

ZorkOne/Command/SpiritProcessor.cs
arsindelve: Now verified against `DEAD-FUNCTION` (`1actions.zil:3113`) and corrected in 1293046. One real bug surfaced: the ZIL `COND` is **order-sensitive** and `RUB` appears in two buckets — the `OPEN/CLOSE/EAT/...` bucket (`:3122`) comes *before* the `TAKE/RUB` bucket (`:3134`), so `rub`/`touch` should return "Even such an action is beyond your capabilities", not "your hand passes through." My original switch had them in the wrong bucket.

`SpiritProcessor` now mirrors the COND order exactly, the buckets are narrowed to the verbs DEAD-FUNCTION actually names (examine/read/etc. correctly fall through to the catch-all), and `LOOK` only adds "objects appear indistinct" when the room has contents (`:3141`). Added a test asserting the ordering.

ZorkOne/Command/SpiritProcessor.cs
arsindelve: Done in 1293046 — `SpiritProcessor` now matches via the shared `Verbs.cs` arrays (`KillVerbs`, `BreakVerbs`, `OpenVerbs`, `CloseVerbs`, `DrinkVerbs`, `TouchVerbs`, `TakeVerbs`, `DropVerbs`, `ThrowVerbs`, `LookVerbs`) instead of hardcoded literals, so the spirit recognizes the same synonyms as the rest of the engine. Only the handful of verbs with no array yet (eat/inflate/turn/burn/wait/score/diagnose/inventory) remain as literals.

ZorkOne/Command/SpiritProcessor.cs
```diff
@@ -0,0 +1,130 @@
...
+            return "You need no light to guide you. ";
+
+        // ATTACK, MUNG, ALARM, SWING.
+        if (Matches(command, Verbs.KillVerbs) || Matches(command, Verbs.BreakVerbs) || Matches(command, "swing", "alarm"))
+            return "All such attacks are vain in your condition. ";
+
+        // OPEN, CLOSE, EAT, DRINK, INFLATE, DEFLATE, TURN, BURN, TIE, UNTIE, RUB (rub/touch live here).
+        if (Matches(command, Verbs.OpenVerbs) || Matches(command, Verbs.CloseVerbs) ||
+            Matches(command, Verbs.DrinkVerbs) || Matches(command, Verbs.TouchVerbs) ||
+            Matches(command, "eat", "inflate", "deflate", "turn", "burn", "tie", "untie"))
```
arsindelve, line 63: Isn't eat in Verbs.cs?

arsindelve:
Thanks — worked through both reviews. Where a point touched ZIL fidelity I cross-checked against `1actions.zil` / `1dungeon.zil` from the pinned historicalsource commit. Summary of what changed vs. what's faithful as-is:

**Acted on (pushed in `79add15`):**

- **`AboveGroundRooms` vs the ZIL table** — real bug, fixed. Diffed against `ABOVE-GROUND` (`1dungeon.zil:2630`). My list wrongly included `ForestFour` and omitted the plain `CLEARING`. It's now the exact eleven rooms `RANDOMIZE-OBJECTS` uses. Worth noting the suggested `CanyonBottom`/`DeepCanyon`/`UpATree` are **not** in the table, so I didn't add them. Also documented the C# name mapping (`BehindHouse` = `EAST-OF-HOUSE`, `Clearing` = `GRATING-CLEARING`, `ClearingBehindHouse` = `CLEARING`) and the treasure-vs-non-treasure scatter split we deliberately simplify (`1actions.zil:4109`).
- **Actor turns skipped on spirit verbs** — added a doc callout confirming it's intentional: the dungeon is "frozen/unreal" to a spirit, NPCs/daemons only advance on movement (which falls through to normal processing), and movement is the only way a spirit dies again — exactly when the permanent-death path matters.
- **`SpiritProcessor` first-word-only / `x` for examine** — addressed in the earlier `Verbs.cs` rework (`1293046`): now whole-verb matching against the shared `Verbs.cs` arrays. `examine`/`x` correctly fall to the catch-all, which is faithful — `DEAD-FUNCTION` has no EXAMINE branch (`1actions.zil:3113`).

**Verified faithful — no change:**

- **`-10` on the permanent (3rd-death) branch** — matches `JIGS-UP`: `<SCORE-UPD -10>` runs *before* the `<NOT <L? ,DEATHS 2>>` → `<FINISH>` check (`1actions.zil:4057`), and the *already-dead* case `<FINISH>`es before `SCORE-UPD` (`:4048`) — which is why my already-dead branch returns before `AddPoints(-10)`. Faithful in both cases.
- **`turn off lamp` asymmetry** — `DEAD-FUNCTION` has no LAMP-OFF branch, so the original also routes turn-off to a generic response; not worth special-casing.
- **`again` while a spirit** — minor edge case, left as-is.

**Doc nit:** the PR body's "10 new tests" is now stale — `SpiritDeathTests.cs` has 13 after the review-driven additions.

Build clean (0 warnings); `ZorkOne.Tests` 1281 + `UnitTests` 934 pass.

arsindelve:
Thanks — all three addressed (pushed in `a1a5b3a`):

1. **Doc-comment vs. actual LOOK text** — real inconsistency, fixed by making the implementation faithful rather than just editing the comment. The ZIL `DEAD-FUNCTION` LOOK prints *both* "The room looks strange and unearthly[, and objects appear indistinct]" *and* — for a room that isn't naturally lit (`<NOT <FSET? ,HERE ,ONBIT>>`) — "Although there is no light, the room seems dimly illuminated." (`1actions.zil:3141`–`3149`). `LookWhileDead` now appends that second clause when the room is an `IDarkLocation { IsNoLongerDark: false }`, so the doc citation and the output now agree. Added a test asserting it at the (dark) Entrance to Hades.

2. **`-10` on the 3rd-death path** — confirmed faithful and now pinned by a test. In `JIGS-UP`, `<SCORE-UPD -10>` runs before the `<NOT <L? ,DEATHS 2>>` → `<FINISH>`, and `<FINISH>` precedes `<RANDOMIZE-OBJECTS>` — so the third death *does* cost 10 points but does *not* scatter. New test `ThirdDeath_DeductsTenPoints_ButDoesNotScatterInventory` locks both halves in. (Only the *already-a-spirit* case skips the point loss, because its `<FINISH>` is above `<SCORE-UPD>`.) Updated the PR description so it no longer implies both permanent branches skip the loss.

3. **Turns as a spirit don't tick the clock** — acknowledged and already documented as intentional in `InterceptPlayerCommand`'s doc comment (the dungeon is "frozen/unreal" to a spirit; only movement advances actors/daemons, and movement is the only way a spirit dies again). Left as-is by design.

Build clean (0 warnings); `ZorkOne.Tests` 1283 + `UnitTests` 934 pass.

arsindelve:
Thanks for the careful trace. All four nits addressed (pushed in `efa1704`):

1. **`AboveGroundRooms` re-resolved per item** — hoisted to a local in `ScatterInventory` before the `foreach`, so the 11-room list is built once per death instead of once per scattered item.
2. **`"turn lamp on"` phrasing** — now handled. Extracted `IsTurnOnLamp`, which accepts the particle either leading (`turn on lamp`, `light lamp`) or trailing the object (`turn lamp on`), mirroring how the ZIL parser collapses all of these to LAMP-ON before DEAD-FUNCTION sees them. Added a test for the trailing form.
3. **`MentionsLamp` substring footgun** — switched from `Contains("lamp")` to a whole-word regex `\b(lamp|lantern)\b`, so a hypothetical future "lampshade" can't be misread as the lamp.
4. **Missing trailing newline** — added at EOF (only `ZorkIContext.cs` was actually missing one by the time I checked; the other two already ended in LF).

Also closed the test-coverage gap you flagged: `PermanentlyDead_SystemCommandsStillWork` asserts a system command (`verbose`) is *not* swallowed by the end-of-game interceptor — confirming the "save/restore/quit still work" claim, which holds because `SystemCommandIntent : GlobalCommandIntent` is consumed at engine step 3, before the step 3b hook.

Build clean (0 warnings); `ZorkOne.Tests` 1284 + `UnitTests` 934 pass.

arsindelve:
Thanks — no code changes needed on this pass; the four observations are all non-blocking and either already handled or intentional:

1. **Simpler spirit parser** — agreed, and intentional: a spirit only needs to recognize a fixed verb table plus movement, so a lightweight prefix matcher (degrading to "You can't even do that.") is the right scope rather than wiring the full AI/regex pipeline into the dead state.
2. **Verb-list overlap (`press`/`destroy`)** — harmless today since the overlapping buckets resolve to the same response; noted for anyone who later splits a bucket.
3. **Frozen clock for stationary spirit verbs** — explicitly documented as a deliberate trade-off (no hunger/thirst daemons in Zork I; only movement re-engages the live world).
4. **Trailing newline on `ZorkIContext.cs`** — already fixed in `efa1704`, which this review predates (it cites 15 tests; HEAD now has 16).

So nothing to push — avoiding a no-op commit that would just restart CI ahead of merge.

PR #319, opened 2026-06-24
Fix #315: scenery mechanism so examining prose-only nouns stops the narrator denying they exist

### LINKED ISSUE #315: Planetfall: Narrator falsely denies existence of scenery objects mentioned in room description (Miniaturization Booth keyboard)
## Bug

In the Miniaturization Booth, the room description explicitly states:

> "Mounted on the wall is a small slot, and next to it **a keyboard with numeric keys**."

However, examining the keyboard returns a narrator response that falsely tells the player it doesn't exist:

```
> examine keyboard
Ah, the legendary invisible keyboard—often mistaken for air. Keep searching, Ensign.
```

And on repeat:
```
> examine keyboard
Ah, the elusive keyboard—it's like trying to find a black hole with a flashlight. Keep searching, Ensign.
```

The narrator actively misleads the player into thinking the keyboard is absent, even though the room description placed it there. The player who tries to understand the booth's interface (a completely natural impulse) is told to "keep searching" for something that the room just told them is right there.

**Repro:**
```
god mode go miniaturization booth
examine keyboard
```

Reproduced on prod.

## Root Cause

The engine has no dedicated `examine keyboard` handler — `quiet "examine keyboard"` returns the "no effect" sentinel, meaning the command falls through entirely to the AI narrator. The narrator has no grounding information about the keyboard (it exists only in the room's prose description, not as a discrete game object), so it generates a "not found / keep looking" style response.

This is a general narrator calibration gap: **when `examine <noun>` has no engine handler but the noun appears verbatim in the room's description text, the narrator should describe the scenery item neutrally, not deny its existence.**

The root failure is that the narrator's fallback behaviour for unhandled `examine` is tuned around "the object isn't here" — correct when the noun genuinely isn't in the room, but wrong when the noun appears in the room description and the player reasonably expects to be able to inspect it.

The original Planetfall source (`planetfall-source/`) lists the keyboard as pure scenery — it is described in the booth's room prose but not a separate game object with its own action routine. The original parser would return a terse "You can't do that" or "That's not something you need to refer to." The AI narrator's "keep searching, Ensign" is a harmful departure: it implies the object is *missing*, not just *non-interactive*.

## Impact

- Any player who tries `examine keyboard` in the Miniaturization Booth is actively lied to.
- The correct interaction (`type 384`) is a non-obvious verb; a player's first instinct is to examine the keyboard to learn what to do. The misleading response increases player frustration.
- The pattern is not unique to the keyboard — any room with scenery nouns that have no engine handler (common throughout both Zork I and Planetfall) is vulnerable to the same false-negative narrator behaviour.

**Severity:** Low — the puzzle remains solvable since `type <number>` still works. But the narrator actively contradicts the room description, which erodes trust in the game's descriptions.

## Suggested Fix

Two complementary approaches:

1. **Short-term / per-object:** Add a minimal examine handler for `keyboard` in the Miniaturization Booth that returns a static string (e.g., "The keyboard consists of numeric keys; you could use it to type in a sector number."). This closes the gap for this room.

2. **Structural / narrator-level:** When the AI narrator receives an `examine <noun>` with no engine response, pass the current room description as additional context. Instruct the narrator: *"If the noun appears in the room description, describe it as scenery (don't claim it's absent)."* This fixes the entire class of false-negative responses across all rooms.

## TDD Plan

In `Planetfall.Tests` (or the appropriate test project per zorkai's `CLAUDE.md`):

```csharp
// Red: examine keyboard in Miniaturization Booth should NOT claim the keyboard is absent
[Test]
public async Task MiniBooth_ExamineKeyboard_DoesNotClaimAbsent()
{
    var response = await SendCommand("examine keyboard");
    // Should NOT contain language suggesting absence
    StringAssert.DoesNotContain("not here", response.Response.ToLower());
    StringAssert.DoesNotContain("keep searching", response.Response.ToLower());
    StringAssert.DoesNotContain("invisible", response.Response.ToLower());
}
```

Green: add the static scenery handler (approach 1) or fix narrator context (approach 2).

---

Found via the AdventureBreaker harness (black-box prod + narrator A/B confirmation). AB-045.

arsindelve:
## What & why

Closes #315. In the Miniaturization Booth (and many other rooms), the prose names things that aren't backed by a game object — "a keyboard with numeric keys". Examining them fell through to the AI narrator, which *falsely denied they existed* ("Ah, the legendary invisible keyboard—often mistaken for air"). The room description even says it's right there, so the narrator was actively contradicting the game's own prose.

Key finding: the narrator **already receives the room description**, so the issue's "no grounding" diagnosis is slightly off — this is a calibration false-negative (the deflection prompt over-denies), not missing context. That reframing led to a deterministic, testable fix instead of a prompt tweak.

## The mechanism

A lightweight, **data-driven** scenery layer (not real `IItem`s):

- **`SceneryItem`** record: `Nouns`, `ExaminationDescription`, optional `CannotBeTakenReason`.
- **`LocationBase.Scenery`** virtual + resolution in `RespondToSimpleInteraction`, matched **after** real items so a genuine object always shadows scenery sharing a noun. `examine` → description; `take` → grounded refusal; any other verb → `NoVerbMatch` ("you can't do that to it") — never "no such thing is here".

Why data, not items: `ItemPlacedHere` dedupes by `GetType()` ([LocationBase.cs](GameEngine/Location/LocationBase.cs)), so a shared data-driven item class would silently drop the 2nd scenery entry per room. Keeping scenery as pure data sidesteps that, needs no serialization, and stays out of hint/available-action lists.

## First consumers

Examine text is an **original paraphrase** of the original ZIL `*-PSEUDO` routines — never verbatim, no invented affordances (per the repo's ZIL-fidelity rules):

- **Planetfall:** Miniaturization Booth (keyboard), Large Office & Balcony (ocean), Waiting Area (benches), Library (carpet), Courtyard & West Wing (castle).
- **Zork I:** North of House (house, boards), End of Rainbow (river).

## Tests (deterministic, no AI)

`SceneryInvariantTests` (Planetfall) and `ZorkSceneryInvariantTests` (in **UnitTests**, which already hosts the Zork I engine harness and references the ZorkOne game assembly) scan *every* room that declares scenery and assert:

1. entries are well-formed (lowercase/trimmed nouns, non-empty description);
2. no duplicate noun across a room's entries;
3. **no scenery noun collides with a real item in the same room** (the main hazard — caught a catalog miss where the Brig already has a real `Graffiti` object);
4. **each noun resolves to its text through the room's real `RespondToSimpleInteraction` override chain** — this calls the method directly, bypassing the test parser (which only knows real-item nouns), and catches the subtle trap of a room override that forgets to call `base`.

Because the gate scans by reflection, it automatically covers future rooms as the sweep continues.

## Reviewer notes

- This is **wave 1** of a larger scenery sweep. A discovery pass (kept out of git) catalogued ~267 candidate scenery nouns across 135 rooms; this PR lands only the high-confidence, ZIL-backed entries that are individually verified. Interactive nouns (leap/flush/cross/slot) are deliberately deferred — they need real items, not inert scenery.
- `LocationBase` change is inert for any room with empty `Scenery` (the default), so existing behavior is unchanged everywhere else.

## Verification

All suites green: **UnitTests 961, ZorkOne.Tests 1306, Planetfall.Tests 2417 — 0 failures.**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ node_modules/
 # (planetfall-source is named to avoid a case-insensitive clash with the C# Planetfall/ project.)
 zork1/
 planetfall-source/
+.scenery-work/
--- a/.gitignore
+++ b/.gitignore
@@ -23,4 +23,3 @@ node_modules/
 # (planetfall-source is named to avoid a case-insensitive clash with the C# Planetfall/ project.)
 zork1/
 planetfall-source/
-.scenery-work/
--- a/Planetfall/Location/Kalamontee/ConferenceRoom.cs
+++ b/Planetfall/Location/Kalamontee/ConferenceRoom.cs
@@ -35,6 +35,13 @@ internal class ConferenceRoom : LocationBase
             $"is {(Door.IsOpen ? "open" : "closed")}. To the north is a small room about the size of a phone booth. ";
     }
 
+    protected override IReadOnlyList<SceneryItem> Scenery =>
+    [
+        new(["conference table", "round conference table", "table"],
+            "A large round conference table dominates the room. Whatever meetings it once hosted are long over. ",
+            "The conference table is far too large to take. ")
+    ];
+
     public override void Init()
     {
         StartWithItem<ConferenceRoomDoor>();
--- a/Planetfall/Location/Kalamontee/ConferenceRoom.cs
+++ b/Planetfall/Location/Kalamontee/ConferenceRoom.cs
@@ -38,7 +38,7 @@ internal class ConferenceRoom : LocationBase
     protected override IReadOnlyList<SceneryItem> Scenery =>
     [
         new(["conference table", "round conference table", "table"],
-            "A large round conference table dominates the room. Whatever meetings it once hosted are long over. ",
+            "It's a large, round conference table. ",
             "The conference table is far too large to take. ")
     ];
 
--- a/Planetfall/Location/Kalamontee/Dorm/SanfacD.cs
+++ b/Planetfall/Location/Kalamontee/Dorm/SanfacD.cs
@@ -1,8 +1,6 @@
-using GameEngine.Location;
-
 namespace Planetfall.Location.Kalamontee.Dorm;
 
-internal class SanfacD : LocationWithNoStartingItems
+internal class SanfacD : SanfacBase
 {
     public override string Name => "Sanfac D";
 
@@ -20,10 +18,4 @@ internal class SanfacD : LocationWithNoStartingItems
                "The fixtures are dry and dusty, the room dead and deserted. You marvel at how " +
                "little the millenia and cultural gulfs have changed toilet bowl design. The only exit is south. ";
     }
-
-    protected override IReadOnlyList<SceneryItem> Scenery =>
-    [
-        new(["fixtures", "fixture"], "The fixtures are bone dry and thick with dust. ",
-            "The fixtures are plumbed into the wall. ")
-    ];
 }
--- a/Planetfall/Location/Kalamontee/Dorm/SanfacD.cs
+++ b/Planetfall/Location/Kalamontee/Dorm/SanfacD.cs
@@ -20,4 +20,10 @@ internal class SanfacD : LocationWithNoStartingItems
                "The fixtures are dry and dusty, the room dead and deserted. You marvel at how " +
                "little the millenia and cultural gulfs have changed toilet bowl design. The only exit is south. ";
     }
+
+    protected override IReadOnlyList<SceneryItem> Scenery =>
+    [
+        new(["fixtures", "fixture"], "The fixtures are bone dry and thick with dust. ",
+            "The fixtures are plumbed into the wall. ")
+    ];
 }
--- a/Planetfall/Location/Shuttle/ShuttleCabin.cs
+++ b/Planetfall/Location/Shuttle/ShuttleCabin.cs
@@ -17,7 +17,7 @@ public abstract class ShuttleCabin : LocationWithNoStartingItems
     protected override IReadOnlyList<SceneryItem> Scenery =>
     [
         new(["seating", "seat", "seats", "freight", "cargo"],
-            "Rows of seating for a couple of dozen passengers, with open space at the back for freight. ",
+            "There is seating for a couple of dozen passengers, with open space at the back for freight. ",
             "The seats are bolted to the cabin floor. ")
     ];
 }
--- a/Planetfall/Location/Shuttle/ShuttleCabin.cs
+++ b/Planetfall/Location/Shuttle/ShuttleCabin.cs
@@ -12,4 +12,12 @@ public abstract class ShuttleCabin : LocationWithNoStartingItems
             "This is the cabin of a large transport, with seating for around 20 people plus space for freight. There " +
             $"are open doors at the eastern and western ends of the cabin, and a doorway leads out to a wide platform to the {Exit}.";
     }
+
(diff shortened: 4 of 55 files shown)


Planetfall/Location/Kalamontee/Dorm/SanfacD.cs
```diff
@@ -20,4 +20,10 @@ protected override string GetContextBasedDescription(IContext context)
                "The fixtures are dry and dusty, the room dead and deserted. You marvel at how " +
```
arsindelve, line 18: Can you put these in the base class for all the Sanfacs?

Planetfall/Location/Kalamontee/ConferenceRoom.cs
```diff
@@ -35,6 +35,13 @@ protected override string GetContextBasedDescription(IContext context)
             $"is {(Door.IsOpen ? "open" : "closed")}. To the north is a small room about the size of a phone booth. ";
     }
 
+    protected override IReadOnlyList<SceneryItem> Scenery =>
+    [
+        new(["conference table", "round conference table", "table"],
+            "A large round conference table dominates the room. Whatever meetings it once hosted are long over. ",
```
arsindelve: Rewrite this. Too flowery. Keep it simple

Planetfall/Location/Shuttle/ShuttleCabin.cs
```diff
@@ -12,4 +12,12 @@ protected override string GetContextBasedDescription(IContext context)
             "This is the cabin of a large transport, with seating for around 20 people plus space for freight. There " +
             $"are open doors at the eastern and western ends of the cabin, and a doorway leads out to a wide platform to the {Exit}.";
     }
+
+    // Shared by both shuttle cars (Alfie and Betty), which inherit this cabin description.
+    protected override IReadOnlyList<SceneryItem> Scenery =>
+    [
+        new(["seating", "seat", "seats", "freight", "cargo"],
+            "Rows of seating for a couple of dozen passengers, with open space at the back for freight. ",
```
arsindelve: Incomplete sentence

ZorkOne/Location/ForestLocation/ForestThree.cs
```diff
@@ -38,4 +38,11 @@ protected override string GetContextBasedDescription(IContext context)
     {
         return "This is a dimly lit forest, with trees all around. ";
     }
+
+    protected override IReadOnlyList<SceneryItem> Scenery =>
+    [
+        new(["trees", "tree", "forest"],
```
arsindelve: Put this in the base class

ZorkOne/Location/AragainFalls.cs
```diff
@@ -32,6 +32,13 @@ protected override string GetContextBasedDescription(IContext context)
                 : "A beautiful rainbow can be seen over the falls and to the west. ");
     }
 
+    protected override IReadOnlyList<SceneryItem> Scenery =>
+    [
+        new(["waterfall", "falls", "aragain falls"],
+            "An enormous waterfall, a good four hundred and fifty feet from lip to river. The spray hangs in the air as a rainbow. ",
```
arsindelve: incomplete sentence

.gitignore
```diff
@@ -23,3 +23,4 @@ node_modules/
 # (planetfall-source is named to avoid a case-insensitive clash with the C# Planetfall/ project.)
 zork1/
 planetfall-source/
+.scenery-work/
```
arsindelve: What's this? If it's leftovers then delete it.

PR #320, opened 2026-06-24
Fix #316: "say Ulysses" (and say-verb synonyms / quoted names) dismisses the cyclops

### LINKED ISSUE #316: "say Ulysses" does not dismiss the cyclops — only bare "Ulysses" is handled
## Bug

In the Cyclops Room, the canonical Zork I solution phrase \`say Ulysses\` has no effect. Only the bare name \`Ulysses\` (or \`Odysseus\`) dismisses the cyclops.

**Repro:**
\`\`\`
> say Ulysses
This action or command has no effect on the game.

> Ulysses
The cyclops, hearing the name of his father's deadly nemesis, flees the room by knocking down the wall on the east of the room.
\`\`\`

Reproduced on prod.

## Root cause

\`ZorkOne/Location/MazeLocation/CyclopsRoom.cs\` — \`RespondToSpecificLocationInteraction\`:

\`\`\`csharp
if (!new List<string> { "ulysses", "odysseus" }.Contains(input.ToLower().Trim())
    || !HasItem<Cyclops>()
    || GetItem<Cyclops>().IsSleeping)
    return await base.RespondToSpecificLocationInteraction(input, context, client);
\`\`\`

The check is against the raw input — \`"say ulysses"\` is not in the list \`{"ulysses", "odysseus"}\`, so it falls through to the base and returns no-effect. The \`say\` verb prefix is never stripped.

In the original Zork I, \`SAY ULYSSES\` is the documented solution and was parsed by the V-SAY routine that ultimately triggered the cyclops-flee routine regardless of whether "say" prefixed the name. Both \`ULYSSES\` and \`SAY ULYSSES\` were accepted.

## Impact

A player using the standard documented phrasing from any Zork I walkthrough or manual (\`say Ulysses\`) gets no feedback — the cyclops simply ignores them. This is a medium-severity parser gap on the main puzzle solution of the Cyclops Room, a room every player must navigate to reach the Treasure Room.

## Suggested fix

Strip a leading \`"say "\` prefix from input before the list check, or extend the list to also match \`"say ulysses"\` and \`"say odysseus"\`:

\`\`\`csharp
var normalized = input.ToLower().Trim();
if (normalized.StartsWith("say ")) normalized = normalized[4..].Trim();
if (!new List<string> { "ulysses", "odysseus" }.Contains(normalized) ...)
\`\`\`

## TDD plan

In \`ZorkOne.Tests/People/CyclopsTests.cs\`, add two test cases alongside the existing \`Ulysses\` / \`Odysseus\` passing cases:

- \`"say Ulysses"\` → cyclops flees (wall knocked down, new exits appear)
- \`"say Odysseus"\` → same result

Red before: both return no-interaction. Green after: both trigger flee.

---
Found via the AdventureBreaker harness (black-box prod + white-box confirmation). AB-046

arsindelve:
## Summary

Fixes #316 — in the Cyclops Room, the canonical Zork I solution phrase `say Ulysses` had no effect. Only the bare name `Ulysses`/`Odysseus` dismissed the cyclops, so a player following any standard walkthrough got a misleading "no effect" message on the room's main puzzle.

## Root cause

`CyclopsRoom.RespondToSpecificLocationInteraction` matched the **raw** input against `{"ulysses", "odysseus"}`. `"say ulysses"` was never in that list, so it fell through to the base handler and returned no-effect. The `say` verb prefix was never stripped.

## Changes

- **Strip a leading say-verb prefix before matching.** Uses `Verbs.SayVerbs` (`say`, `speak`, `utter`, `shout`, `yell`, `whisper`, `recite`, …) as the single source of truth rather than a hardcoded `"say "`.
- **Strip wrapping quotes** (straight and curly, single and double) so quoted forms like `say "Ulysses"` and `"Ulysses"` also work.

Both `Ulysses` and `Odysseus` continue to work bare, as before.

## Tests (TDD)

Added to `ZorkOne.Tests/People/CyclopsTests.cs`:
- `SayUlysses_Response`, `SayOdysseus_Response` — the `say <name>` forms
- `SayUlyssesInQuotes_Response`, `UlyssesInQuotes_Response` — quoted forms
- `SayVerbSynonyms_DismissTheCyclops` — parameterized over `speak`/`utter`/`shout`/`whisper`/`recite` (quoted and unquoted)

Red confirmed before the fix (no-effect instead of flee), green after. Full `ZorkOne.Tests` suite: **1312 passed, 0 failed**.

In the original Zork I, `SAY ULYSSES` is the documented solution; the V-SAY routine triggered the cyclops-flee regardless of whether `say` prefixed the name. This restores that behavior.

### DIFF
--- a/ZorkOne.Tests/People/CyclopsTests.cs
+++ b/ZorkOne.Tests/People/CyclopsTests.cs
@@ -51,6 +51,39 @@ public class CyclopsTests : EngineTestsBase
     }
 
     [Test]
+    public async Task SayUlysses_Response()
+    {
+        // Issue #316: the canonical walkthrough phrasing "say Ulysses" must also dismiss the
+        // cyclops, not just the bare name "Ulysses".
+        var target = GetTarget();
+        target.Context.CurrentLocation = Repository.GetLocation<CyclopsRoom>();
+        target.Context.ItemPlacedHere(Repository.GetItem<Torch>());
+
+        var response = await target.GetResponse("say Ulysses");
+
+        response.Should()
+            .Contain(
+                "The cyclops, hearing the name of his father's deadly nemesis, " +
+                "flees the room by knocking down the wall on the east of the room");
+    }
+
+    [Test]
+    public async Task SayOdysseus_Response()
+    {
+        // Issue #316: "say Odysseus" must also dismiss the cyclops.
+        var target = GetTarget();
+        target.Context.CurrentLocation = Repository.GetLocation<CyclopsRoom>();
+        target.Context.ItemPlacedHere(Repository.GetItem<Torch>());
+
+        var response = await target.GetResponse("say Odysseus");
+
+        response.Should()
+            .Contain(
+                "The cyclops, hearing the name of his father's deadly nemesis, " +
+                "flees the room by knocking down the wall on the east of the room");
+    }
+
+    [Test]
     public async Task Asleep_Odysseus_Response()
     {
         var target = GetTarget();
--- a/ZorkOne.Tests/People/CyclopsTests.cs
+++ b/ZorkOne.Tests/People/CyclopsTests.cs
@@ -104,6 +104,8 @@ public class CyclopsTests : EngineTestsBase
     [TestCase("shout Ulysses")]
     [TestCase("whisper Odysseus")]
     [TestCase("recite \"Ulysses\"")]
+    [TestCase("say 'Ulysses'")]
+    [TestCase("say \"Odysseus\"")]
     public async Task SayVerbSynonyms_DismissTheCyclops(string command)
     {
         // Issue #316: any Verbs.SayVerbs synonym (not just "say") should prefix the name.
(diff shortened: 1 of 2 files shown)


arsindelve:
Addressed the `"tell"` flag from the review: `"tell"` is now excluded from the speak-verb prefix in `CyclopsRoom`, mirroring `ConversationHandler.NamelessSpeechVerbs`, so `"tell ulysses"` no longer dismisses the cyclops. Also switched the prefix match to the suggested `FirstOrDefault` form and added a `Tell_IsNotTreatedAsASpeakVerb` test. The `ZorkOneGlobalCommandFactory` easter-egg note was out of scope for #316, so left as-is.

arsindelve:
Closed the test-coverage gap noted in the review: added `say 'Ulysses'` (single-quote stripping, which the code supports but nothing exercised) and a quoted+prefixed `say "Odysseus"` case. The duplication and `StripNonChars` style notes are noted as optional follow-ups and left as-is per the review.

PR #324, opened 2026-06-25
Fix #323: library terminal pressing 0 from leaf works after HTTP round-trip

### LINKED ISSUE #323: Library terminal: pressing 0 from a text entry doesn't navigate back to submenu (Parent reference lost in serialization)
## Bug

After navigating to any text entry (leaf node) in the library computer terminal, pressing **0** (which the text instructs as "Tiip zeeroo tuu goo tuu aa hiiyur levul" = "Type zero to go to a higher level") does nothing — the terminal feeps and shows "typing that character has no meaning at the moment" instead of returning to the parent submenu.

**Repro (narrator OFF):**
```
turn on terminal
press 1          → History submenu shown
press 2          → "Graat Hiiaatus" text shown; ends with "Tiip zeeroo tuu goo tuu aa hiiyur levul."
press 0          → "The terminal feeps, and a message briefly appears on the screen explaining that
                    typing that character has no meaning at the moment."
                    ↑ WRONG — should show the History submenu again
```

Reproduced deterministically on prod. Affects every text entry in every submenu (History, Culture, Technology, Geography, Project, Games — all tested).

**Expected:** `press 0` at a leaf text node navigates back to the parent submenu, showing the submenu menu again.

**Actual:** NoEffect message — the terminal feeps and nothing changes.

The text the player just read explicitly instructs them to press 0, making this a hard contradiction (instruction says press 0, engine rejects it).

## Root cause

 — :

```csharp
internal string GoUp()
{
    if (CurrentItem.Parent is null)   // ← always true for leaf nodes after an HTTP round-trip
        return NoEffect;

    CurrentItem = CurrentItem.Parent;
    return $"The screen clears and a different menu appears:\n\n{CurrentItem.Text}";
}
```

The  of a leaf node is null because all menu classes (e.g. , ) use **computed expression-body properties** for :

```csharp
// Planetfall/Item/Lawanda/Library/Computer/HistoryMenu.cs
internal override List<MenuItem> Children =>
[
    new() { Children = null, Parent = this, Text = MenuOne },
    new() { Children = null, Parent = this, Text = MenuTwo },
    new() { Children = null, Parent = this, Text = MenuThree }
];
```

Every access to  creates a brand-new list of brand-new  objects.  calls  three times in one invocation (null-check, count-check, index access) — each call returns a different list. The leaf assigned to  is from the third list; its  field holds the correct menu-class instance **in-process**. However, when the game state is **serialized to JSON** between HTTP calls, the serializer follows the object graph:  →  (menu class instance) →  (another computed call → new items) → their  (another menu class) → … This creates a deep or circular reference tree that the serializer either truncates or represents without back-pointers. After deserialization,  is null.

The bug is that  from a submenu works correctly (submenu-class types override  with , so Parent is always non-null) but leaf  instances rely on their  backing field surviving serialization — which it doesn't.

*(ZIL not available locally; the intended behavior is described by the in-game text that instructs "Tiip zeeroo tuu goo tuu aa hiiyur levul" — pressing 0 must navigate up.)*

## Impact

Every player who reads any text entry in the library terminal is stuck: the only exit is to turn off the terminal and restart it (resetting to the main menu). The terminal's own instructions are contradicted. Severity: medium — library terminal is important for understanding the game's lore and plot; the navigation being broken means players effectively can only read one entry per power cycle.

## Suggested fix

Rebuild the menu tree once (e.g. in  constructor or a static factory), wiring  and  as true object-graph references **not** computed on every access, so they survive JSON serialization. Alternatively, store only a path (e.g. a stack of indices) in  and recompute  from the root on each call — no Parent references needed.

A minimal approach: make  store a breadcrumb (e.g. ), navigate by walking from the root + path, and drop the runtime / reference model entirely.

## TDD plan

In , add:
- Red:  asserts the response contains submenu text (not the feep/NoEffect message).
- Green: passes after the fix.

Also add similar tests for each submenu (Culture, Technology, Geography, Project, Games).

---

Found via the AdventureBreaker harness (black-box prod + white-box C# confirmation). AB-051

arsindelve:
## Bug

Pressing `0` from any leaf text node in the library computer terminal returned the feep/NoEffect message instead of navigating back to the parent submenu. The terminal's own on-screen text tells the player to press 0, making this a hard contradiction.

Affected every text entry in every submenu (History, Culture, Technology, Geography, Project, Games). Players were effectively stuck: the only exit was to turn off the terminal and restart.

## Root cause

`MenuItem.Parent` is `internal`, so Newtonsoft.Json silently skips it during serialization. After an HTTP round-trip (each web request serializes then deserializes full game state), the deserialized `CurrentItem` has `Parent = null`. `MenuState.GoUp()` checked `CurrentItem.Parent is null` and returned `NoEffect`. In-process (same memory, no serialization) it worked fine, which is why the existing `GoIntoChildMenuAndBackToSubmenu` tests passed.

## Fix

Replaced the Parent-reference model with a **breadcrumb path** (`List<int> Path`) in `MenuState`:

- `Path` is a `public List<int>` — serialized and restored correctly across round-trips.
- `CurrentItem` is now a `[JsonIgnore]` computed property that walks from `new MainMenu()` following the path indices on every access.
- `GoDown(n)` appends to `Path`; `GoUp()` pops from `Path`.
- No changes to `MenuItem`, `MainMenu`, `HistoryMenu`, or any other menu class.

## Proof (red → green)

`ComputerTerminalTests.PressZero_FromLeafNode_AfterSaveAndRestoreRoundTrip_ReturnsSubmenuText` — navigates to a leaf, calls `SaveGame()` + `RestoreGame()` to simulate the HTTP round-trip, then presses `0` and asserts the response contains `HistoryMenu.MainMenu` (not `NoEffect`).

- **Before fix:** FAIL — response contained the feep message.
- **After fix:** PASS.

All 2 418 `Planetfall.Tests` + `UnitTests` tests pass.

Closes #323.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/Planetfall/Item/Lawanda/Library/Computer/MenuState.cs
+++ b/Planetfall/Item/Lawanda/Library/Computer/MenuState.cs
@@ -11,42 +11,68 @@ public class MenuState
     internal const string ReachedLowestLevel =
         "\"Yuu hav reect xe loowist levul uv xe liibreree indeks. Pleez tiip zeeroo tuu goo tuu aa hiiyur levul. If yuu reekwiir asistins, kawl xe liibrereein.\"";
 
-    [UsedImplicitly] public MenuItem CurrentItem { get; set; } = new MainMenu();
+    // Breadcrumb path: 1-based child indices from the root. Empty = at MainMenu.
+    // Storing the path (not the MenuItem object) is what survives JSON serialization —
+    // MenuItem.Parent is internal so Newtonsoft.Json skips it, making CurrentItem.Parent
+    // null after a round-trip and breaking GoUp() (issue #323).
+    [UsedImplicitly] public List<int> Path { get; set; } = [];
+
+    // Cached result of walking Path from the root. Invalidated whenever Path mutates.
+    [Newtonsoft.Json.JsonIgnore] private MenuItem? _current;
+
+    // Computed from Path; cached so repeated accesses within one method call don't
+    // re-allocate a fresh child list at every level.
+    [Newtonsoft.Json.JsonIgnore]
+    public MenuItem CurrentItem => _current ??= BuildCurrentItem();
+
+    private MenuItem BuildCurrentItem()
+    {
+        MenuItem current = new MainMenu();
+        foreach (var index in Path)
+        {
+            var children = current.Children;
+            if (children is null || index < 1 || index > children.Count)
+                return current;
+            current = children[index - 1];
+        }
+        return current;
+    }
 
     /// <summary>
     /// Navigates to the parent menu item if it exists, otherwise indicates that the action is invalid.
     /// </summary>
-    /// <returns>Returns a message indicating the result of the navigation or invalid action.</returns>
     internal string GoUp()
     {
-        if (CurrentItem.Parent is null)
-            return
-                NoEffect;
+        if (Path.Count == 0)
+            return NoEffect;
 
-        CurrentItem = CurrentItem.Parent;
+        Path.RemoveAt(Path.Count - 1);
+        _current = null;
         return $"The screen clears and a different menu appears:\n\n{CurrentItem.Text}";
     }
 
     /// <summary>
     /// Navigates to a child menu item based on the provided index, if valid, or indicates the action is invalid.
     /// </summary>
-    /// <param name="menuItem">The index of the desired child menu item.</param>
-    /// <returns>Returns a message indicating the result of the navigation or invalid action.</returns>
     internal string GoDown(int menuItem)
     {
-        if (CurrentItem.Children is null)
-            return
-                ReachedLowestLevel;
+        var children = CurrentItem.Children;
 
-        if (menuItem > CurrentItem.Children.Count)
-            return
-                NoEffect;
+        if (children is null)
+            return ReachedLowestLevel;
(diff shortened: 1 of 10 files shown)


Planetfall/Item/Lawanda/Library/Computer/MenuState.cs
```diff
@@ -11,42 +11,66 @@ public class MenuState
...
 
-    [UsedImplicitly] public MenuItem CurrentItem { get; set; } = new MainMenu();
+    // Breadcrumb path: 1-based child indices from the root. Empty = at MainMenu.
+    // Storing the path (not the MenuItem object) is what survives JSON serialization —
+    // MenuItem.Parent is internal so Newtonsoft.Json skips it, making CurrentItem.Parent
+    // null after a round-trip and breaking GoUp() (issue #323).
+    [UsedImplicitly] public List<int> Path { get; set; } = [];
+
+    // Cached result of walking Path from the root. Invalidated whenever Path mutates.
+    [Newtonsoft.Json.JsonIgnore] private MenuItem? _current;
```
arsindelve, line 21: add a using to reduce this length

PR #325, opened 2026-06-25
Follow-up #324: fix Path setter cache staleness and align [JsonIgnore] style

arsindelve:
Two small follow-ups from the [PR #324](https://github.com/arsindelve/ZorkAI/pull/324) review thread.

## Changes

**Cache staleness** — `MenuState.Path` had a plain auto-property setter. If `Path` were ever assigned directly on an already-navigated instance (the JSON deserializer does this on a fresh instance where `_current` is already `null`, so no live bug today), the cached `_current` would not be invalidated. Added a backing field `_path` with an explicit setter that nulls `_current` on every assignment, closing the latent footgun.

**Style alignment** — `[Newtonsoft.Json.JsonIgnore]` was fully qualified. Every other file in the codebase (`Context.cs`, `PlanetfallContext.cs`, etc.) uses `using Newtonsoft.Json;` + bare `[JsonIgnore]`. Aligned to match.

### DIFF
--- a/Planetfall.Tests/ComputerTerminalTests.cs
+++ b/Planetfall.Tests/ComputerTerminalTests.cs
@@ -502,4 +502,19 @@ public class ComputerTerminalTests : EngineTestsBase
             "pressing 0 from a leaf should navigate up to the parent submenu, not feep");
         response.Should().NotContain(MenuState.NoEffect);
     }
+
+    [Test]
+    public void MenuState_Should_InvalidateCurrentItemCache_When_PathIsDirectlyReassigned()
+    {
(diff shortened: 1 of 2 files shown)


arsindelve:
Added  in  — creates a `MenuState`, warms the cache via `GoDown(1)`, then directly reassigns `Path = [2]` and asserts `CurrentItem` is no longer the same instance.

arsindelve:
Leaving the assertion as-is — the test's purpose is to prove the setter nulls the backing field (reference invalidation), not to test BuildCurrentItem's correctness (covered elsewhere). Noted for future tests where value identity matters more.

PR #326, opened 2026-06-26
Fix #220: implement V-OIL — oiling a living Floyd thanks you

### LINKED ISSUE #220: Planetfall: can't oil Floyd (V-OIL not implemented)
Oiling Floyd is a small flavor interaction from the original that the C# port is missing. `Planetfall/Item/Kalamontee/Mech/OilCan.cs` has no verb handling at all, and there is no `oil`/`lubricate` verb anywhere — so `oil floyd` currently falls through to the generic "you can't do that" path instead of Floyd's signature thank-you.

**Grammar** (`planetfall-source/syntax.zil:446-448`):
```
<SYNTAX OIL OBJECT = V-OIL>
<SYNTAX OIL OBJECT WITH OBJECT (HAVE) = V-OIL>
<SYNONYM OIL LUBRICATE>
```

**`V-OIL`** (`planetfall-source/verbs.zil:1738-1757`):
- **No indirect object** (`oil floyd`): if the oil can is in inventory, it redirects to `OIL <thing> WITH OIL-CAN`; otherwise → *"Oil it with what?"*
- **Indirect object is the oil can** (`oil floyd with oil can`, or the redirect above):
  - target is **Floyd and Floyd is alive** → *"Floyd thanks you for your thoughtfulness."*
  - any other target → *"The <thing> doesn't need oiling."*
- **Indirect object is anything else** → *"You can't use a/an <thing> as an oil can!"*

"Floyd is alive" is `FSET? FLOYD RLANDBIT`. On Floyd this flag means alive — it's cleared at his death in the Bio Lab (`compone.zil:1938`: `<FCLEAR ,FLOYD ,RLANDBIT>`). So oiling **dead** Floyd gives "doesn't need oiling," not the thanks.

**POUR → OIL redirect** (`OIL-CAN-F`, `compone.zil:1358-1367`): pouring with the oil can redirects to `V-OIL`, so `pour oil can on floyd` also yields the thank-you. (`V-POUR` itself just forbids pouring non-liquids — the redirect lives on the oil can's own action routine.)

## How this maps onto the C# engine

Floyd's other flavor verbs already follow exactly the pattern this needs — `FloydSocialResponses.HandleSocialInteraction` (`Planetfall/Item/Kalamontee/Mech/FloydPart/FloydSocialResponses.cs`) matches verbs like `play`/`kiss`/`kick`/`rub`/`kill` and returns a `FloydConstants` string. It is:
- gated on `floyd.IsOn` (the `ONBIT` analog), and
- only reached when Floyd is **not** dead — `Floyd.RespondToSimpleInteraction` returns to base before social handling when `HasDied` is true (`Floyd.cs:210-211`).

That maps cleanly to the ZIL alive-check: implement oiling here and a dead/off Floyd naturally won't give the thank-you.

Multi-noun (`oil floyd with oil can`) routes through `Floyd.RespondToMultiNounInteraction` (`Floyd.cs:294`).

## Recommended scope

**Core (do this):**
1. Add `FloydConstants.Oil = "Floyd thanks you for your thoughtfulness. "`.
2. In `FloydSocialResponses.HandleSocialInteraction`, add `if (action.Match(["oil", "lubricate"], floyd.NounsForMatching)) return new PositiveInteractionResult(FloydConstants.Oil);` — handles `oil floyd` / `lubricate floyd`.
3. Optionally add an `OilVerbs = ["oil", "lubricate"]` array to `Model/Verbs.cs` for consistency with the existing `Verbs.KillVerbs` usage, and reference it instead of the inline literal.

**Fidelity decisions to make explicit (each is a small add):**
- **Require holding the oil can?** ZIL does: `oil floyd` with no can → *"Oil it with what?"*. Recommended: match it — check `context.HasItem<OilCan>()`, else return "Oil it with what? ". (If we'd rather keep it a pure social verb like kiss/kick with no item check, that's the cheaper alternative — call it out in the PR.)
- **`oil floyd with oil can` (multi-noun):** the ZIL grammar explicitly supports `OIL OBJECT WITH OBJECT`. Recommended: handle it in `Floyd.RespondToMultiNounInteraction` (or normalize "... with oil can" down to the simple form) so it gives the same thank-you.
- **`pour oil can on/over floyd`:** optional — mirrors the ZIL POUR→OIL redirect; lower priority.
- **Generic targets** (`oil <non-Floyd>` → "doesn't need oiling"; `oil X with <non-can>` → "you can't use … as an oil can!"): these are verb-level behaviors not tied to Floyd. Optional/deferrable; the issue's core is the Floyd flavor.

**Heads-up for implementation:** `OilCan.NounsForMatching` already includes `"oil"`, so confirm the parser treats `oil` as the verb in `oil floyd` (not as a reference to the can). The existing kiss/kick/rub flavor verbs prove arbitrary verbs reach `RespondToSimpleInteraction`, so this should route the same way — the proving test below verifies it.

## TDD plan (per CLAUDE.md, bug fixes require a failing-first test)

Add to `Planetfall.Tests/FloydTests.cs`, mirroring `KissFloyd`/`RubFloyd` (deterministic; `Take<OilCan>()` forces the can into inventory regardless of where it normally starts, as `SpoolReaderTests` does with `Take<Pliers>()`):

```csharp
[Test]
public async Task OilFloyd()
{
    var target = GetTarget();
    StartHere<RobotShop>();
    GetItem<Floyd>().IsOn = true;
    Take<OilCan>();                       // only needed if we require holding the can

    var response = await target.GetResponse("oil floyd");

    response.Should().Contain("thoughtfulness");
}
```
Before the fix this fails (no "thoughtfulness" — generic fallback); after, it passes. Add companions as scope dictates: `lubricate floyd`, `oil floyd with oil can`, and — if implementing the faithful no-can branch — `oil floyd` without the can → "Oil it with what?".

## Acceptance criteria

- [ ] `oil floyd` (and `lubricate floyd`) while Floyd is on/alive → "Floyd thanks you for your thoughtfulness."
- [ ] `oil floyd with oil can` → same response
- [ ] Dead/off Floyd does **not** give the thank-you (falls through, matching the RLANDBIT check)
- [ ] Existing Floyd interactions unchanged; new failing-first test added and passing; `dotnet test Planetfall.Tests` green

## Affected files

- `Planetfall/Item/Kalamontee/Mech/FloydPart/FloydConstants.cs` — add `Oil` string
- `Planetfall/Item/Kalamontee/Mech/FloydPart/FloydSocialResponses.cs` — match `oil`/`lubricate`
- `Planetfall/Item/Kalamontee/Mech/FloydPart/Floyd.cs` — multi-noun `oil … with oil can` (if in scope)
- `Model/Verbs.cs` — optional `OilVerbs` array
- `Planetfall.Tests/FloydTests.cs` — proving test(s)

_ZIL claims in the original issue verified: `V-OIL` at `verbs.zil:1738-1757`, POUR→OIL redirect at `compone.zil:1358-1367`. Found via the ZIL behavior audit._

arsindelve:
Closes #220.

Planetfall was missing the original's oil-Floyd flavor interaction. There was no `oil`/`lubricate` verb anywhere, so `oil floyd` fell through to the generic "you can't do that" path instead of Floyd's signature thank-you.

## What this does

Ports the `V-OIL` behavior from the original ZIL (`planetfall-source/verbs.zil:1738-1757`, grammar `syntax.zil:446-448`):

- **`oil floyd` / `lubricate floyd` while holding the oil can** → *"Floyd thanks you for your thoughtfulness."*
- **`oil floyd` with no oil can in hand** → *"Oil it with what?"* — the original prompts for the instrument rather than oiling (faithful to the no-indirect-object branch, which only redirects to the thank-you when the can is in inventory).
- **`oil floyd with oil can` (multi-noun)** → same thank-you, gated on the `with` preposition (the ZIL grammar is `OIL OBJECT WITH OBJECT`) **and only when the can is actually held** — the `(HAVE)` flag requires the indirect object be in inventory, so a can merely sitting in the room does not count.
- **Dead/off Floyd never gives the thank-you** — this falls out naturally: `Floyd.RespondToSimpleInteraction` returns to base before social handling when `HasDied`, and the multi-noun path is gated on `IsOn && !HasDied`. This matches the ZIL `RLANDBIT` ("alive") check, which is cleared at Floyd's death in the Bio Lab.

## How it maps onto the engine

This follows the exact pattern Floyd's other flavor verbs (`kiss`/`kick`/`rub`) already use in `FloydSocialResponses.HandleSocialInteraction` — gated on `floyd.IsOn` and returning a `FloydConstants` string. The `oil`/`lubricate` verbs are kept **local to `FloydSocialResponses`** (an internal static array) rather than added to the shared `Model/Verbs.cs` thesaurus, since they're Floyd-specific — mirroring how the other flavor verbs are matched. The handler also now receives `IContext` so the oil branch can check whether the can is held.

## Deferred (explicitly optional per #220, not covered here)

These are called out as optional/deferrable in the issue and are **not** implemented in this PR — left for a follow-up:
- Generic targets: `oil <non-Floyd>` → "doesn't need oiling", and `oil X with <non-can>` → "you can't use … as an oil can!" (verb-level behaviors not tied to Floyd).
- `pour oil can on/over floyd` (the ZIL `OIL-CAN-F` POUR→OIL redirect).

## TDD

Per CLAUDE.md (bug fixes require a failing-first test), added to `Planetfall.Tests/FloydTests.cs`:

- `OilFloyd` — holding the can → thank-you
- `LubricateFloyd` — synonym → thank-you
- `OilFloyd_WithOilCan` — multi-noun, can held → thank-you
- `OilFloyd_WithOilCan_NotHeld_DoesNotThank` — can in room, not held → no thank-you (the `(HAVE)` flag)
- `OilFloyd_WithoutOilCan` — no can → "Oil it with what?"
- `OilFloyd_Dead_DoesNotThank` — dead Floyd (with `IsOn = true`) falls through, exercising the `HasDied` gate

Each behavioral test failed before its corresponding fix and passes after. The deterministic `TestParser` and `base-mappings.json` gain the new verbs so the test parser routes them, mirroring the existing `play with floyd` precedent.

## Verification

- `dotnet test Planetfall.Tests` → **2430 passed, 0 failed**
- `dotnet test UnitTests` → **961 passed, 0 failed**

## Files

- `Planetfall/Item/Kalamontee/Mech/FloydPart/FloydConstants.cs` — `Oil` + `OilWithWhat` strings
- `Planetfall/Item/Kalamontee/Mech/FloydPart/FloydSocialResponses.cs` — local `OilVerbs` array; match `oil`/`lubricate` with can-in-hand check
- `Planetfall/Item/Kalamontee/Mech/FloydPart/Floyd.cs` — pass context; multi-noun `oil … with oil can` (held-only, `with`-gated)
- `Planetfall.Tests/FloydTests.cs` — proving tests
- `UnitTests/TestParser.cs`, `UnitTests/IntentMappings/base-mappings.json` — test-parser verb/mapping support

### DIFF
--- a/Model/Verbs.cs
+++ b/Model/Verbs.cs
@@ -76,4 +76,10 @@ public static class Verbs
     ///     ones (e.g. Up A Tree). Centralized here so every location recognizes the same synonyms.
     /// </summary>
     public static readonly string[] JumpVerbs = ["jump", "leap"];
+
+    /// <summary>
+    ///     The "oil it" family — applying oil from the oil can. Centralized so Floyd's V-OIL flavor
+    ///     interaction (oiling a living Floyd thanks you) recognizes the same synonyms.
+    /// </summary>
+    public static readonly string[] OilVerbs = ["oil", "lubricate"];
 }
--- a/Model/Verbs.cs
+++ b/Model/Verbs.cs
@@ -76,10 +76,4 @@ public static class Verbs
     ///     ones (e.g. Up A Tree). Centralized here so every location recognizes the same synonyms.
     /// </summary>
     public static readonly string[] JumpVerbs = ["jump", "leap"];
-
-    /// <summary>
-    ///     The "oil it" family — applying oil from the oil can. Centralized so Floyd's V-OIL flavor
-    ///     interaction (oiling a living Floyd thanks you) recognizes the same synonyms.
-    /// </summary>
-    public static readonly string[] OilVerbs = ["oil", "lubricate"];
 }
--- a/Planetfall.Tests/FloydTests.cs
+++ b/Planetfall.Tests/FloydTests.cs
@@ -185,6 +185,22 @@ public class FloydTests : EngineTestsBase
     }
 
     [Test]
+    public async Task OilFloyd_WithOilCan_NotHeld_DoesNotThank()
+    {
+        var target = GetTarget();
+        StartHere<RobotShop>();
+        GetItem<Floyd>().IsOn = true;
+        // The oil can is in the room but never picked up. The ZIL grammar is
+        // OIL OBJECT WITH OBJECT (HAVE) — the (HAVE) flag requires the can be held, so this
+        // must NOT give the thank-you even though the can resolves in scope.
+        GetLocation<RobotShop>().ItemPlacedHere(GetItem<OilCan>());
+
+        var response = await target.GetResponse("oil floyd with oil can");
+
+        response.Should().NotContain("thoughtfulness");
+    }
+
+    [Test]
     public async Task OilFloyd_WithoutOilCan()
     {
         var target = GetTarget();
--- a/Planetfall.Tests/FloydTests.cs
+++ b/Planetfall.Tests/FloydTests.cs
@@ -220,6 +220,10 @@ public class FloydTests : EngineTestsBase
         var floyd = GetItem<Floyd>();
         floyd.HasDied = true;
         floyd.HasEverBeenOn = true;
+        // IsOn = true so the only thing blocking the thank-you is the HasDied gate in
+        // Floyd.RespondToSimpleInteraction (which returns to base before social handling).
+        // Without this, the !IsOn early-return in HandleSocialInteraction would mask the HasDied path.
+        floyd.IsOn = true;
         Take<OilCan>();
 
         var response = await target.GetResponse("oil floyd");
--- a/Planetfall.Tests/FloydTests.cs
+++ b/Planetfall.Tests/FloydTests.cs
@@ -9,6 +9,7 @@ using Moq;
 using Planetfall.Item;
 using Planetfall.Item.Feinstein;
 using Planetfall.Item.Kalamontee.Admin;
+using Planetfall.Item.Kalamontee.Mech;
 using Planetfall.Item.Kalamontee.Mech.FloydPart;
 using Planetfall.Location.Kalamontee;
 using Planetfall.Location.Kalamontee.Mech;
@@ -142,6 +143,72 @@ public class FloydTests : EngineTestsBase
         var response = await target.GetResponse("rub floyd");
 
         response.Should().Contain("contented sigh");
+    }
+
+    [Test]
+    public async Task OilFloyd()
+    {
+        var target = GetTarget();
+        StartHere<RobotShop>();
+        GetItem<Floyd>().IsOn = true;
+        Take<OilCan>();
+
+        var response = await target.GetResponse("oil floyd");
+
+        response.Should().Contain("thoughtfulness");
+    }
+
+    [Test]
+    public async Task LubricateFloyd()
+    {
+        var target = GetTarget();
+        StartHere<RobotShop>();
+        GetItem<Floyd>().IsOn = true;
+        Take<OilCan>();
+
+        var response = await target.GetResponse("lubricate floyd");
+
+        response.Should().Contain("thoughtfulness");
+    }
+
+    [Test]
+    public async Task OilFloyd_WithOilCan()
+    {
+        var target = GetTarget();
+        StartHere<RobotShop>();
+        GetItem<Floyd>().IsOn = true;
+        Take<OilCan>();
+
+        var response = await target.GetResponse("oil floyd with oil can");
+
+        response.Should().Contain("thoughtfulness");
+    }
+
+    [Test]
+    public async Task OilFloyd_WithoutOilCan()
+    {
+        var target = GetTarget();
+        StartHere<RobotShop>();
+        GetItem<Floyd>().IsOn = true;
+
+        var response = await target.GetResponse("oil floyd");
+
+        response.Should().Contain("Oil it with what?");
+    }
+
+    [Test]
+    public async Task OilFloyd_Dead_DoesNotThank()
+    {
+        var target = GetTarget();
+        StartHere<RobotShop>();
+        var floyd = GetItem<Floyd>();
+        floyd.HasDied = true;
+        floyd.HasEverBeenOn = true;
+        Take<OilCan>();
+
+        var response = await target.GetResponse("oil floyd");
+
+        response.Should().NotContain("thoughtfulness");
     }
 
     [Test]
--- a/Planetfall/Item/Kalamontee/Mech/FloydPart/Floyd.cs
+++ b/Planetfall/Item/Kalamontee/Mech/FloydPart/Floyd.cs
@@ -324,6 +324,9 @@ public class Floyd : QuirkyCompanion, IAmANamedPerson, ICanHoldItems, ICanBeGive
         // grammar is OIL OBJECT WITH OBJECT (HAVE) — the (HAVE) flag requires the can be HELD, so
         // gate on context.HasItem<OilCan>() (matching the single-noun branch), not merely
         // GetItemInScope, which would also resolve a can sitting in the room.
+        // Both checks are needed and not redundant: GetItemInScope(NounTwo) is OilCan verifies the
+        // *named* indirect object is the can (so "oil floyd with sword" is rejected even while a can
+        // is held), and HasItem<OilCan>() enforces the (HAVE) flag (the can must be in inventory).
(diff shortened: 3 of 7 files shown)


arsindelve:
Thanks — good catch. Confirmed the bug: `Repository.GetItemInScope` searches the room before inventory, so `oil floyd with oil can` thanked you even with the can left in the room, contradicting the `(HAVE)` flag.

Fixed in 850621a:
- Gated the multi-noun branch on `context.HasItem<OilCan>()` (mirroring the single-noun branch).
- Added the missing TDD coverage — `OilFloyd_WithOilCan_NotHeld_DoesNotThank` (can placed in the room, never taken) fails before the fix and passes after.
- Noted the explicitly-optional deferred items (`oil <non-Floyd>`, `oil X with <non-can>`, `pour oil can on floyd`) in the PR description so they aren't assumed covered.

`dotnet test Planetfall.Tests` → 2430 passed, 0 failed.

---

arsindelve:
Thanks for the thorough pass. Addressed the two actionable items in 59babc6:

- **Test gap (`OilFloyd_Dead_DoesNotThank`)**: now sets `IsOn = true`, so it exercises the `HasDied` gate in `Floyd.RespondToSimpleInteraction` directly instead of being masked by the `!IsOn` early-return.
- **"Redundant" dual check**: kept both checks (it's the second case you identified — `GetItemInScope(NounTwo) is OilCan` validates the *named* indirect object so `oil floyd with sword` is rejected even when a can is held; `HasItem<OilCan>()` enforces the `(HAVE)` flag) and added a comment explaining why both are needed.

On the other notes:
- **Fallback asymmetry** (can in room, not held → generic fallback vs. "Oil it with what?"): left intentional. The strictly faithful response here would be the parser's "you aren't holding the oil can" from the `(HAVE)` flag, which isn't modeled in this engine; "Oil it with what?" belongs to the distinct no-indirect-object grammar line. Falling through rather than borrowing a message from the other form avoids inventing a divergence — flagging it as a deliberate small gap, deferrable with the other optional V-OIL behaviors.
- **`oil` verb / `"oil"` noun overlap**: verified the parser resolves positionally — `oil floyd` routes as verb (covered by the passing `OilFloyd` test); no confusion with `take oil`.

`dotnet test Planetfall.Tests` → 2430 passed, 0 failed.

---

Model/Verbs.cs
```diff
@@ -76,4 +76,10 @@ public static class Verbs
     ///     ones (e.g. Up A Tree). Centralized here so every location recognizes the same synonyms.
     /// </summary>
     public static readonly string[] JumpVerbs = ["jump", "leap"];
+
+    /// <summary>
+    ///     The "oil it" family — applying oil from the oil can. Centralized so Floyd's V-OIL flavor
+    ///     interaction (oiling a living Floyd thanks you) recognizes the same synonyms.
+    /// </summary>
+    public static readonly string[] OilVerbs = ["oil", "lubricate"];
```
arsindelve: This is very specific and should actually not live in VERBS.cs

arsindelve: Agreed — moved it out of `Model/Verbs.cs` in 5162cfa. `OilVerbs` now lives in `FloydSocialResponses` as an internal static array, matching how the other Floyd-only flavor verbs (play/kiss/kick/rub) are matched locally rather than in the shared thesaurus. Both the single-noun handler and the multi-noun branch reference the one definition.

---

arsindelve:
Thanks for the review. Adopted the optional preposition-tightening in 5162cfa: the multi-noun branch now calls `action.MatchPreposition(["with"])`, so connectors like "using"/"near" no longer trigger the thank-you — faithful to the `OIL OBJECT WITH OBJECT` grammar. (Same commit also moves `OilVerbs` out of the shared `Verbs.cs` into `FloydSocialResponses`, per the repo owner's inline note.)

On the `OilFloyd_WithOilCan_NotHeld_DoesNotThank` nit: kept it asserting `NotContain("thoughtfulness")` rather than pinning the exact fallback. The not-held case falls through to the base multi-noun handler, whose response can be AI-generated, so asserting its precise text would make the test non-deterministic/brittle — the absence of the thank-you is the behavior under test.

`dotnet test Planetfall.Tests` → 2430 passed; `UnitTests` → 961 passed.

---

PR #347, opened 2026-07-01
Fix #342: block taking room items in the dark via the TakeIntent path

### LINKED ISSUE #342: Zork: room objects can still be taken in pitch darkness
## Bug

In Zork I prod, room objects can still be taken while the room is pitch black.

Live Attic repro through AdventureBreaker:

```text
> turn off lantern
It is pitch black. You are likely to be eaten by a grue.

> take rope
Taken.
```

The same repro works in quiet / engine-only mode, so this is not narrator flavor. The structured response also correctly hides exits/actions in the dark, but the actual take command still mutates inventory.

This is a current prod regression of closed #7 ("Should not be able to take items in a dark room").

## Root cause

The ordinary simple-interaction path has a dark-room guard:

```csharp
if (context.ItIsDarkHere && !refersToALightSource)
    return (null, "It's too dark to see! ");
```

`GameEngine/IntentEngine/SimpleInteractionEngine.cs:38-44`

But take/drop is handled by the separate `TakeOrDropInteractionProcessor`. `TakeIt` checks portability, existing inventory, closed containers, and load, then calls `context.Take(castItem)` without checking darkness:

```csharp
if (!context.HaveRoomForItem(castItem))
    return new PositiveInteractionResult("Your load is too heavy. ");

context.Take(castItem);
context.RememberAntecedentNoun(castItem.NounsForMatching.FirstOrDefault());
```

`GameEngine/Item/ItemProcessor/TakeOrDropInteractionProcessor.cs:189-224`

That means a direct noun match such as `take rope` can still resolve an in-room object and move it into inventory even though the current location is dark.

## Original behavior (ZIL - ground truth)

The original parser only searches the room's visible object list when `LIT` is true, then always searches the player inventory. If an object is not found and the room is dark, it reports darkness rather than resolving the room object:

```zil
<COND (,LIT
       <FCLEAR ,PLAYER ,TRANSBIT>
       <DO-SL ,HERE ,SOG ,SIR>
       <FSET ,PLAYER ,TRANSBIT>)>
<DO-SL ,PLAYER ,SH ,SC>
```

`zork1/gparser.zil:1074-1078`

For not-found visible-room nouns in darkness:

```zil
<COND (<OR ,LIT <VERB? TELL ;WHERE ;WHAT ;WHO>> ...)
      (T <TELL "It's too dark to see!" CR>)>
```

`zork1/gparser.zil:1122-1132`

So taking the rope from the Attic while the lantern is off should fail with a darkness response, not succeed.

## Impact

Medium-low severity parser/state-integrity regression. It does not block progression, but it bypasses the core Zork darkness rule and lets players manipulate unseen room objects while the game claims it is pitch black.

## Suggested fix

Add a dark-room visibility guard to the take path before resolving/taking in-room objects:

- Allow taking/dropping carried inventory as appropriate.
- Allow interacting with a carried light source enough to turn it on via the existing simple path.
- Do not allow `TakeIt` to take an object whose current location is the dark room unless that object is already in inventory / otherwise visible under the original rules.

A focused fix likely belongs in `TakeOrDropInteractionProcessor.GetItemsToTake` or `TakeIt`, near the point where the chosen item is known and before `context.Take(castItem)`.

## TDD plan

Add deterministic Zork I tests:

1. Put the player in Attic with lantern on and rope present.
2. Turn off lantern.
3. Run `take rope`.
4. Assert response contains `too dark` / darkness failure and inventory does not contain rope.
5. Add a control that `turn on lantern` still works in darkness and `take rope` succeeds after relighting.

Found via the AdventureBreaker harness (black-box prod + white-box/ZIL confirmation). AB-045.

arsindelve:
## Summary

Fixes #342. In production, the real AI parser tags a bare command like `take rope` as a `TakeIntent`, which `GameEngine.cs` dispatches directly to `TakeOrDropInteractionProcessor.Process(TakeIntent, ...)`. That path never checked `context.ItIsDarkHere`, so a player could `take` a room object while the room was pitch black (e.g. turning off the lantern in the Attic and still taking the rope) — a regression of the original `SimpleIntent` darkness guard in `SimpleInteractionEngine`, which only covers verb+noun input the AI parser tags generically rather than as an explicit take/drop intent.

- `TakeOrDropInteractionProcessor.TakeIt` now returns `"It's too dark to see! "` when the room is dark and the resolved item isn't already in the player's inventory, mirroring the original ZIL parser (only resolves room-visible objects when lit; inventory items remain accessible regardless of light).
- Items already carried are unaffected (falls through to the existing "You already have that!" case).

## Root cause

- `GameEngine/GameEngine.cs` routes `TakeIntent`/`DropIntent` straight to `TakeOrDropInteractionProcessor`, bypassing `SimpleInteractionEngine`'s darkness guard (`GameEngine/IntentEngine/SimpleInteractionEngine.cs:43`).
- `TakeOrDropInteractionProcessor.TakeIt` (`GameEngine/Item/ItemProcessor/TakeOrDropInteractionProcessor.cs`) called `context.Take(castItem)` without any darkness check.
- The existing test suite couldn't catch this because `TestParser` (used by all engine-level tests) always resolves `take <noun>` to a `SimpleIntent`, never a `TakeIntent` — so the buggy `TakeIntent`-facing code path had zero test coverage.

## Test plan

- Added `TakeItem_InDarkRoom_ViaTakeIntent_SaysTooDarkAndDoesNotTakeIt` in `UnitTests/SingleNounProcessors/TakeProcessorTests.cs`, which invokes `TakeOrDropInteractionProcessor.Process(TakeIntent, ...)` directly (the same call GameEngine makes for a live AI-tagged "take") against a real `Attic`/`Rope` Repository setup with no light source. Confirmed it fails before the fix ("Taken." / rope in inventory) and passes after.
- Added a control test, `TakeItem_ViaTakeIntent_SucceedsAfterRelightingLantern`, verifying the same path still succeeds once the lantern is turned back on.
- [x] `dotnet test UnitTests` — 981 passed, 1 skipped
- [x] `dotnet test ZorkOne.Tests` — 1725 passed
- [x] `dotnet test Planetfall.Tests` — 2841 passed
- [x] `dotnet test EscapeRoom.Tests` — 7 passed
- [x] `dotnet build` (full solution) — succeeded

---
_Generated by [Claude Code](https://claude.ai/code/session_01F18JTxpasRuobYe9LBC4LW)_

### DIFF
--- a/GameEngine/Item/ItemProcessor/TakeOrDropInteractionProcessor.cs
+++ b/GameEngine/Item/ItemProcessor/TakeOrDropInteractionProcessor.cs
@@ -191,6 +191,14 @@ public class TakeOrDropInteractionProcessor : IVerbProcessor
         if (castItem is null)
             return new NoNounMatchInteractionResult();
 
+        // Issue #342: TakeIntent (a live AI "take" tag) is dispatched here directly from GameEngine,
+        // bypassing the darkness guard that SimpleIntent goes through in SimpleInteractionEngine. Mirror
+        // the original parser, which only resolves room-visible objects when lit - an item already in
+        // inventory stays takeable (falls through to the "already have that" case below) since the player
+        // can always feel what they're carrying regardless of light.
+        if (context.ItIsDarkHere && !context.Items.Contains(castItem))
+            return new PositiveInteractionResult("It's too dark to see! ");
+
         if (!string.IsNullOrEmpty(castItem.CannotBeTakenDescription))
         {
             ((ItemBase)castItem).OnFailingToBeTaken(context);
--- a/GameEngine/StaticCommand/Implementation/TakeEverythingProcessor.cs
+++ b/GameEngine/StaticCommand/Implementation/TakeEverythingProcessor.cs
@@ -17,6 +17,12 @@ public class TakeEverythingProcessor : IGlobalCommand
         Runtime runtime
         )
     {
+        // Issue #342 follow-up: "take all" is a global command matched before the AI parser ever
+        // runs, so none of the TakeIntent/SimpleIntent darkness guards apply to it. Gate here too,
+        // matching the established per-global-command convention (see LookProcessor.LookAround).
+        if (context.ItIsDarkHere)
+            return "It's too dark to see! ";
+
         var items = ((ICanContainItems)context.CurrentLocation).GetAllItemsRecursively;
 
         if (!items.Any(s => s is ICanBeTakenAndDropped || !string.IsNullOrEmpty(s.CannotBeTakenDescription)))
--- a/GameEngine/StaticCommand/Implementation/TakeEverythingProcessor.cs
+++ b/GameEngine/StaticCommand/Implementation/TakeEverythingProcessor.cs
@@ -72,6 +72,15 @@ public class TakeEverythingProcessor : IGlobalCommand
                 continue;
             }
 
+            // Issue #342 follow-up: a live AI TakeIntent that resolves multiple nouns (e.g. "take rope
+            // and knife") reaches this method directly instead of TakeIt, so it needs the same darkness
+            // guard - an item not already in inventory must not be resolvable while the room is dark.
+            if (context.ItIsDarkHere && !context.Items.Contains(item))
+            {
+                sb.AppendLine($"{noun}: It's too dark to see! ");
+                continue;
+            }
+
             // Check if the item is actually available in scope (room or inventory containers)
             var itemsInLocation = ((ICanContainItems)context.CurrentLocation).GetAllItemsRecursively;
             var itemsInInventory = context.GetAllItemsRecursively;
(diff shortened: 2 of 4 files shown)


arsindelve:
Thanks for the thorough pass. On the "possible inconsistency" point: the specific example ("take lamp" while already holding it) isn't actually a fork — `lamp` resolves to `IAmALightSourceThatTurnsOnAndOff`, so `SimpleInteractionEngine`'s `refersToALightSource` carve-out already bypasses its dark check for that noun regardless of verb, landing on the same "You already have that!" as the `TakeIntent` path. The fork is real for a non-light-source already-held item ("take sword" while holding it, dark), though.

Leaving `TakeIt`'s inventory carve-out as-is rather than removing it: issue #342's own "Suggested fix" section explicitly asks for this — "Allow taking/dropping carried inventory as appropriate" and "unless that object is already in inventory" — so this isn't a new interpretation, it's what was requested, and it matches the ZIL rule the issue itself cites (parser always searches inventory regardless of `LIT`).

I'm not making `SimpleInteractionEngine`'s blanket check symmetric in this PR, though — that guard is shared by every verb (examine, open, read, wear, push, ...), not just take/drop, so relaxing it is a materially larger change than what #342 asked for, and I'd want to verify each affected verb's original behavior individually rather than bundle a cross-cutting change into this fix. Recommend filing that as its own follow-up if it's worth pursuing.

The `refersToALightSource` gap in the new guard: agreed it's the same class of divergence, but per your own note nothing in current content reaches it (no unlit light source starts on the floor of a dark room), so adding dead code for it now would violate the repo's "don't design for hypothetical requirements" convention. Happy to add it the moment a room actually needs it.

---

PR #357, opened 2026-07-02
Disarm Planetfall's explosion clock on god-mode teleport

arsindelve:
## Summary

Investigation of #356 found the reported "unconditional death at move 14 silently wipes session progress" is **not a bug** — it's Planetfall's intentional, already-tested escape-pod timed puzzle. `Planetfall.Tests/ExplosionTests.cs` already has `Experience_StayOnDeckNine` (and four sibling tests) asserting that a player who never reaches the escape pod dies exactly as described — faithfully recreating the original Feinstein-explosion sequence. The issue's suggested fix (treat Deck Nine as permanently safe past move 12) would have broken that existing, deliberate mechanic and let players camp indefinitely at the very start of the game.

However, digging into the reproduction turned up a real, narrower problem: **`god mode go <place>`** — the dev/testing teleport command — is a raw `CurrentLocation` swap. It skips `DeckNine.OnLeaveLocation` and `EscapePod.AfterEnterLocation`, so a tester who teleports away from Deck Nine (e.g. to check on later-game content) never trips the hook that would normally unregister `ExplosionCoordinator`. The clock keeps counting turns in the background and can unconditionally kill the tester — wiping their god-mode session via `RestartAfterDeath` — once `Moves` rolls into its move-10-to-14 death window, regardless of what they're actually testing.

## Changes

- Added `Model/Interface/IGodModeTeleportAware`, a game-agnostic marker interface (mirroring the existing `IResettableClockContext` / `ISurvivalClockContext` god-mode affordances) with a single `OnGodModeTeleport()` hook.
- `GodModeProcessor.Go()` now calls this hook after relocating the player via `god mode go`.
- `PlanetfallContext` implements it to remove `ExplosionCoordinator` from the actor list. Games without a comparable location-blind death clock (Zork I) don't implement the interface and are unaffected.

## Test plan

- [x] Added `GodModeGo_DisarmsExplosionClock_SoTeleportedTesterSurvives` to `Planetfall.Tests/ExplosionTests.cs`, following TDD: confirmed red first (`DeathCounter` became 1 — the tester was killed and the session reset), then green after the fix.
- [x] `dotnet test Planetfall.Tests` — 2842 passed, 0 failed (including all existing explosion/death-sequence tests, confirming the intentional Deck-Nine death mechanic itself is untouched).
- [x] `dotnet test UnitTests` — 985 passed, 0 failed (game-agnostic `GodModeProcessor` change, ZorkOne's context doesn't implement the new interface).
- [x] `dotnet test ZorkOne.Tests` — 1750 passed, 0 failed.

https://claude.ai/code/session_01NRyG8MF31eey55kCyYrLjQ

---
_Generated by [Claude Code](https://claude.ai/code/session_01NRyG8MF31eey55kCyYrLjQ)_

### DIFF
--- a/GameEngine/StaticCommand/Implementation/GodModeProcessor.cs
+++ b/GameEngine/StaticCommand/Implementation/GodModeProcessor.cs
@@ -108,6 +108,14 @@ public class GodModeProcessor : IGlobalCommand
         // StartWithItem<T> (a bare Items.Add with no dedupe) would end up with their starting items
         // duplicated in the room.
         context.CurrentLocation = location;
+
+        // Issue #356 follow-up: this is a raw location swap, not a real move - it skips
+        // OnLeaveLocation/AfterEnterLocation, so a game-specific location-blind death clock (e.g.
+        // Planetfall's Feinstein explosion) never gets the chance to disarm itself. Let the context
+        // do that explicitly for a god-mode teleport.
+        if (context is IGodModeTeleportAware teleportAware)
+            teleportAware.OnGodModeTeleport();
+
         return $"Welcome to {location.Name}";
     }
 
--- a/Model/Interface/IGodModeTeleportAware.cs
+++ b/Model/Interface/IGodModeTeleportAware.cs
@@ -1,0 +1,17 @@
+namespace Model.Interface;
+
+/// <summary>
+///     Marker interface for game contexts that run early-game, location-blind timed-death mechanics
+///     (e.g. Planetfall's Feinstein explosion clock) which only check where the player currently is,
+///     not how they got there. "god mode go &lt;place&gt;" is a raw location swap for testing/debugging -
+///     it does not run the normal OnLeaveLocation/AfterEnterLocation hooks - so a tester who teleports
+///     away from the danger zone never trips the hook that would otherwise disarm the clock. It keeps
+///     counting turns in the background and can silently kill the tester once the move count rolls
+///     into its death window, far from whatever they were actually testing. Games that implement this
+///     get a chance to disarm those clocks when a god-mode teleport happens. Games without such
+///     mechanics do not implement this.
+/// </summary>
+public interface IGodModeTeleportAware
+{
+    void OnGodModeTeleport();
+}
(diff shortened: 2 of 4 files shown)


arsindelve:
Pushed a follow-up commit addressing two correctness gaps found in review:

1. **Incomplete coverage** — `OnGodModeTeleport()` only disarmed `ExplosionCoordinator`, leaving `EscapePod`'s own post-landing sinking timer (`YerSinking`/`TurnsAfterStanding`) armed. That timer has the identical location-blind death bug (no `CurrentLocation` check), so a tester who stood out of the safety web and then teleported away could still be silently killed by the sinking pod, several turns later, far from the ship.
2. **Overshooting fix** — the original disarm logic fired unconditionally on every "god mode go X", including teleports *into* Deck Nine/Escape Pod. That meant a developer who teleported in specifically to observe the Feinstein-explosion sequence found it permanently defused on arrival, with no way to trigger it short of restarting the engine.

`OnGodModeTeleport()` now also disarms `EscapePod`, and skips disarming a clock when the teleport destination is the location that clock actually governs — so "god mode go deck nine" (or "... escape pod") still lets you test the sequence, while teleporting *away* correctly disarms both clocks.

Added two TDD tests (`GodModeGo_DisarmsEscapePodSinkingClock_SoTeleportedTesterSurvives`, `GodModeGo_IntoDeckNine_DoesNotDisarmExplosionClock`) proving each was red before the fix and green after. Full suite: `Planetfall.Tests` 2844/2844, `UnitTests` 985/985, `ZorkOne.Tests` 1750/1750, no regressions.

---

arsindelve:
Confirmed and fixed — good catch. `ExplosionCoordinator`'s move-14 case really has no location guard, so my `EscapePod` exception was wrong for the reason you gave: unlike `DeckNine` (where staying armed is intentional design), the pod is never supposed to reach move 14 with the coordinator still armed under normal play, since `EscapePod.AfterEnterLocation` always disarms it first.

Dropped the `EscapePod` exception from the `ExplosionCoordinator` removal check — only `DeckNine` keeps it now. Added your suggested test (`GodModeGo_IntoEscapePod_DoesNotGetKilledByExplosionClock`), confirmed red before the fix, green after. Full suite still passes (2845/985/1750, no regressions).

---

PR #359, opened 2026-07-02
Planetfall: implement magnet's card-scrambling hazard

### LINKED ISSUE #211: Planetfall: magnet doesn't scramble/de-magnetize access cards
The magnet's hazard mechanic — corrupting your access cards while you carry it — isn't implemented. (Its useful function, pulling the steel key out of the crevice, works fine.)

`I-MAGNET` (`planetfall-source/compone.zil:1595-1612`) is a daemon, enabled when you take the magnet (`compone.zil:1572`). Each turn the magnet is in your inventory it sets `SCRAMBLEDBIT` on the **first** carried access card it finds, in this priority order:

```
KITCHEN, SHUTTLE, TELEPORTATION, UPPER-ELEVATOR, LOWER-ELEVATOR, MINI, ID
```

(one card per turn; carry the magnet long enough and they all get corrupted). When the magnet is no longer held, the daemon disables itself. A scrambled card is rejected by its slot with the standard wrong-card sign: *"A sign flashes \"Inkorekt awtharazaashun kard...akses deeniid.\""* (`WRONG-CARD`, `globals.zil:1438`).

## C# current state

`Planetfall/Item/Kalamontee/Mech/Magnet.cs` is an inert `ItemBase, ICanBeTakenAndDropped` — no daemon, no scramble. There is no scrambled state anywhere (`grep -i scrambl Planetfall/` is empty), and the card slots don't check for one. The slot machinery already exists: `SlotBase` (`Planetfall/Item/Kalamontee/Admin/SlotBase.cs`) handles slides and already emits the exact "Inkorekt awtharazaashun kard…" message for a wrong card — so the rejection text is in place.

## Fix

1. **Scrambled state:** add `bool Scrambled` to the `AccessCard` base (`Planetfall/Item/AccessCard.cs`). (Verify `IdCard` derives from `AccessCard`; if it doesn't, include it explicitly so the ID card can scramble too.)
2. **Magnet daemon:** make `Magnet` an `ITurnBasedActor` (register on take, deregister on drop — mirror how other "while X" actors register/unregister). In `Act`, while the magnet is held, scramble carried access cards. The ZIL marks one per turn in the priority order above; replicating that is the faithful choice, though scrambling all currently-held cards in one pass is an acceptable simplification (note whichever you choose).
3. **Slot rejection:** in `SlotBase`, when the player slides the **correct** card type but it is `Scrambled`, return the "Inkorekt awtharazaashun kard...akses deeniid." rejection instead of calling `OnSuccessfulSlide`.
4. **Permanence:** in the original, scrambling is permanent (no un-scramble) — decide whether to match that or offer recovery, and document the choice.

## TDD plan (per CLAUDE.md)

Deterministic (no RNG needed — the daemon is unconditional while held):
- Take the magnet, hold a `KitchenAccessCard`, advance a turn → `card.Scrambled == true`.
- Then swipe that card at the `KitchenSlot` → response is the "Inkorekt awtharazaashun kard…" rejection, and the slot does **not** perform its success action.
- Control: the same card swiped **without** ever carrying the magnet → works normally.
Place tests near the existing card/slot or magnet tests.

## Acceptance criteria

- [ ] Carrying the magnet scrambles carried access cards (priority order or all-held, documented)
- [ ] A scrambled card is rejected by its slot with the "Inkorekt awtharazaashun kard…" message; an unscrambled card still works
- [ ] Dropping the magnet stops further scrambling
- [ ] Deterministic tests added, red-before/green-after; `dotnet test Planetfall.Tests` green

## Affected files

- `Planetfall/Item/AccessCard.cs` — `Scrambled` flag
- `Planetfall/Item/Kalamontee/Mech/Magnet.cs` — scramble daemon (take/drop registration)
- `Planetfall/Item/Kalamontee/Admin/SlotBase.cs` — reject a scrambled (otherwise-correct) card
- `Planetfall.Tests/…` — proving tests

_Verified against `planetfall-source/compone.zil:1595-1612` (`I-MAGNET`) and `globals.zil:1438` (`WRONG-CARD`). Found via the ZIL routine-level behavior audit (no scramble logic exists)._

arsindelve:
## Summary

Fixes #211. The magnet's hazard mechanic — corrupting carried access cards — was never implemented (only its useful function, pulling the steel key from the crevice, worked).

- `Planetfall/Item/AccessCard.cs` — adds a `Scrambled` flag to the `AccessCard` base.
- `Planetfall/Item/Kalamontee/Mech/Magnet.cs` — `Magnet` is now an `ITurnBasedActor`. It registers itself when taken (`OnBeingTaken`) and, each turn it's still held, scrambles one not-yet-scrambled carried access card. **Note:** the original's I-MAGNET (`compone.zil:1595-1612`) walks a fixed type-priority order (Kitchen, Shuttle, Teleportation, Upper Elevator, Lower Elevator, Miniaturization); per review feedback this port deliberately simplified that away — `Act()` just asks `context.Items.OfType<AccessCard>().FirstOrDefault(c => !c.Scrambled)`, so cards scramble in **acquisition order**, not the ZIL's fixed priority. This is an explicitly-acceptable simplification per the issue, and it means any future `AccessCard` subtype is covered automatically with no list to maintain. When dropped (`CurrentLocation` no longer the player), it deregisters itself so scrambling stops.
- `Planetfall/Item/Kalamontee/Admin/SlotBase.cs` — a scrambled card of the *correct* type is now rejected instead of opening its door. **Deliberate deviation from the original:** the ZIL's WRONG-CARD (`globals.zil:1438`) gives a scrambled card the identical "incorrect authorization" message as an actual wrong-type card. Since scrambling is silent (no notice when it happens) and permanent, that left the player no way to tell "wrong card, try another" apart from "your card is ruined, stop carrying it with the magnet" — we judged that ambiguity unfair, so a scrambled card now gets its own distinct message (`"Damejd kard...akses deeniid."`, in the game's established phonetic style) instead of matching the original verbatim.
- Scrambling itself is permanent — no un-scramble — matching the original.
- `IdCard` deliberately stays out of `AccessCard`: `Floyd.cs:395-397` already depends on `IdCard` not being an `AccessCard` (it's enumerated explicitly alongside three `AccessCard` subtypes, by design), and there is no ID card slot in this port for a scrambled ID to matter against. Documented with a comment on `IdCard`.

No walkthrough changes were needed — the existing walkthroughs always `drop all` (dropping the magnet) well before picking up any access card, so they never exercised the hazard.

## Test plan

- [x] Tests in `Planetfall.Tests/AdminCorridorSouthTests.cs`: taking the magnet while holding a card scrambles it; not carrying the magnet never scrambles; one card per turn in acquisition order; dropping the magnet stops further scrambling.
- [x] Tests in `Planetfall.Tests/CardTests.cs`: a scrambled card is rejected at its own slot with its own distinct message (door stays closed); an unscrambled card still opens the door; the scrambled-card message differs from the wrong-card-type message.
- [x] Verified all new tests fail (red) against the pre-fix code, then pass (green) after each fix.
- [x] `dotnet test Planetfall.Tests` — 2853/2853 passing.
- [x] `dotnet test UnitTests` — 985/985 passing (1 skipped, pre-existing).
- [x] `dotnet build Zork.sln` — full solution builds clean.

https://claude.ai/code/session_015EcDc2Ng8AySY589yv7Psw


Planetfall/Item/Kalamontee/Mech/Magnet.cs
```diff
@@ -18,4 +22,48 @@ public override string GenericDescription(ILocation? currentLocation)
...
+    /// </summary>
+    public Task<string> Act(IContext context, IGenerationClient client)
+    {
+        if (CurrentLocation != context)
+        {
+            context.RemoveActor(this);
+            return Task.FromResult(string.Empty);
+        }
+
+        _ = TryScramble<KitchenAccessCard>(context)
```
arsindelve: Surely this can be refactored

arsindelve: Fixed in f141c95 — replaced the `||` short-circuit chain with a `Func<IContext, bool>[]` array walked in a plain loop that stops at the first card it scrambles. Same behavior, clearer control flow.

---

Planetfall/Item/Kalamontee/Mech/Magnet.cs
```diff
@@ -18,4 +22,59 @@ public override string GenericDescription(ILocation? currentLocation)
...
+    }
+
+    /// <summary>
+    /// I-MAGNET's priority order (compone.zil:1595-1612): checked in this order each turn, the first
+    /// not-yet-scrambled carried card is the one that gets corrupted.
+    /// </summary>
+    private static readonly Func<IContext, bool>[] ScramblePriorityOrder =
+    [
+        TryScramble<KitchenAccessCard>,
+        TryScramble<ShuttleAccessCard>,
```
arsindelve: This is still fragile and cumbersome. Just get all types of IAccessCard.

arsindelve: Fixed in b40153b — dropped the per-type dispatch entirely. Since `AccessCard` is the shared base, `Act()` now just does `context.Items.OfType<AccessCard>().FirstOrDefault(c => !c.Scrambled)` and scrambles that. No list of types to maintain, and any future `AccessCard` subtype is covered automatically. This trades the original's fixed type-priority order for acquisition order, which the issue explicitly allows as a simplification — updated the affected test accordingly.

---

PR #361, opened 2026-07-02
Fix #354: score/look/inventory/time no longer consume a turn

### LINKED ISSUE #354: Meta-verbs (score/look/inventory) incorrectly consume a Planetfall turn, advancing the survival clock into unwarranted deaths
## Bug

In Planetfall, `score`, `look`, and `inventory` — meta/informational verbs that should be free actions — each advance the turn counter and the in-game survival clock exactly like a real action (`take`, `open`, movement).

Reproduced on prod, in an isolated fresh session (narrator OFF, ruling out narrator involvement):

```
moves=0 time=4558   (game start)
> score      -> moves=1 time=4671  (+54)
> look       -> moves=2 time=4725  (+54)
> inventory  -> moves=3 time=4779  (+54)
> score      -> moves=4 time=4833  (+54)
```

Same result with narrator ON.

**Concrete observed harm:** replaying the engine's own verified golden walkthrough (from the `Planetfall.Tests` NUnit fixtures, which call `score` 7 times before a certain point — completely normal player behavior) command-for-command against live prod, those 7 "free" checks silently burn 7 extra turns (378 extra clock ticks) the golden path never accounts for. The very next scripted command (moving W into the Large Office, which the golden path expects to be safe) instead triggers an unavoidable sleep/hunger death (`...you awake as several ferocious beasts...You have died`) that does not occur in the verified walkthrough. A player who does nothing wrong beyond periodically checking their score/inventory/surroundings can be killed by survival-clock drift entirely outside their control.

## Root cause

`score`, `look`, and `inventory` are registered as **global commands** rather than **system commands**:

`GameEngine/StaticCommand/GlobalCommandFactory.cs:30-79` — `InventoryProcessor`, `LookProcessor`, and `ScoreProcessor` are all returned from `GetGlobalCommands`, alongside genuinely turn-consuming verbs like movement and `wait`. The engine already has a separate turn-free bucket — `GetSystemCommands` (same file, lines 94-131) — used for `save`/`restore`/`quit`/`verbose`/`restart`, whose own doc comment (lines 85-91) says these "should always work, no matter the state of the game." `score`/`look`/`inventory` fit that description but were never routed there.

`GameEngine/GameEngine.cs:405-423` makes the split explicit:

```csharp
// 3. ------- System, or "meta" commands - like save, restore, quit, verbose etc. Does not count as a turn. No actor or turn processing.
var systemCommand = _parser.DetermineSystemIntentType(playerInput);
...
// Everything below here counts as a turn. Pre-process the turn.
// See if the context needs to notify us of anything. Are we sleepy? Hungry?
var contextPrepend = Context.ProcessBeginningOfTurn();
```

Because `score`/`look`/`inventory` aren't recognized as system commands, they fall through to `Context.ProcessBeginningOfTurn()` (`GameEngine/Context.cs:419-423`, `Moves++`), then get dispatched at `GameEngine/GameEngine.cs:480-493` — explicitly commented "Global commands... **These DO count as a turn**. We must process actors afterwards" — through `ProcessActorsAndContextEndOfTurn` (`GameEngine/GameEngine.cs:779-812`), which runs NPC actors and calls `Context.ProcessEndOfTurn()`.

For Planetfall specifically, `PlanetfallContext.ProcessEndOfTurn()` (`Planetfall/PlanetfallContext.cs:241-246`) unconditionally advances the survival clock:

```csharp
public override string? ProcessEndOfTurn()
{
    Repository.GetItem<Chronometer>().CurrentTime += TurnTimeIncrement; // 54
    ...
}
```

and `PlanetfallContext.ProcessBeginningOfTurn()` (lines ~150-238) is also where hunger/sleep escalation and death checks fire — so checking your score or inventory can, on its own, push the clock past a sleep/hunger threshold and trigger a death.

## Expected behavior

Meta/informational verbs (`score`, `look`, `inventory`) are conventionally free actions in Infocom-style parsers — they let the player check state without spending game time. This repo's own architecture already encodes exactly that distinction (the system-command vs. global-command split above); `score`/`look`/`inventory` are simply on the wrong side of it. (Note: a full ZIL ground-truth checkout wasn't available to cite an exact original-source routine for this issue — the divergence is established here from the engine's own internal design intent and the golden-walkthrough regression above, not from `planetfall-source/`.)

## Impact

**Severity: High.** Silent, cumulative, and entirely outside player control — every `score`/`look`/`inventory` check quietly steals a turn and 54 clock ticks. In a survival-clock game like Planetfall this can turn ordinary, cautious play (checking your status before acting) into an unwarranted death, and desyncs the engine's own verified walkthrough fixtures from live prod behavior.

## Suggested fix

Give `score`/`look`/`inventory` (and likely the existing `time`/`current time` query, which has the same shape) a "free" classification that skips `Context.ProcessBeginningOfTurn()`, actor processing, and `Context.ProcessEndOfTurn()` — mirroring how system commands already bypass turn processing in `ProcessSingleSentence`. A minimal approach: introduce a marker (e.g. `IFreeGlobalCommand : IGlobalCommand`) implemented by `InventoryProcessor`, `LookProcessor`, `ScoreProcessor`, and `CurrentTimeProcessor`, and check for it in `GameEngine.ProcessSingleSentence` (`GameEngine/GameEngine.cs:480-493`) to route those to the same no-turn path system commands use, without disturbing the existing step 4 (location-specific raw interaction) ordering that some rooms rely on for custom `look` behavior (e.g. `ZorkOne/Location/MazeLocation/TreasureRoom.cs`, `ZorkOne/Location/Temple.cs`, `Planetfall/Location/Computer/Station384.cs`, which already call `LookProcessor` directly for their own purposes and would be unaffected).

## TDD plan

Add a deterministic test (real `Repository`/`Context`, no AI, no randomness) in `Planetfall.Tests` (see `EngineTestsBase.cs` for the full-pipeline `GameEngine<PlanetfallGame, PlanetfallContext>` harness pattern) asserting that after `score`, `look`, and `inventory`, both `Context.Moves` and `Repository.GetItem<Chronometer>().CurrentTime` are unchanged from their pre-command values. Confirm it's red today (each command currently advances `Moves` by 1 and `CurrentTime` by 54), then apply the minimal fix and confirm green. Re-run `Planetfall.Tests` in full to make sure no location that depends on `look`/`score` counting as a turn regresses.

---

Found via the AdventureBreaker harness (black-box prod + white-box engine confirmation). `AB-046`

arsindelve:
## Summary

Fixes #354. `score`, `look`, `inventory`, and `time`/`current time` were registered as ordinary global commands, so every one of these informational checks advanced `Context.Moves` and — for time-based games like Planetfall — ran `Context.ProcessBeginningOfTurn()`/`ProcessEndOfTurn()`, which advances the survival clock (`Chronometer.CurrentTime += 54`) and runs hunger/sleep escalation and death checks. Routine status checks could silently push the clock past a sleep/hunger threshold and cause an unwarranted death.

- Added an `IFreeGlobalCommand : IGlobalCommand` marker interface (`Model/Interface/IGlobalCommand.cs`), implemented by `InventoryProcessor`, `LookProcessor`, `ScoreProcessor`, and `CurrentTimeProcessor`.
- `GameEngine.ProcessSingleSentence` now classifies the input *before* `Context.ProcessBeginningOfTurn()` runs, and skips both `ProcessBeginningOfTurn()` and `ProcessEndOfTurn()` for free commands.
- Actor processing (`ProcessActors()`) still runs for free commands, so the rest of the world keeps moving — chase-scene monsters, countdown timers (e.g. the Cryo-Elevator), Floyd, NPCs, etc. This split is deliberate and is confirmed against the game's own golden-walkthrough fixtures (e.g. `WalkthroughTestOne.cs`'s Cryo-Elevator countdown, and `WalkthroughMutantChase.cs`'s `RandomPause` test), which rely on "look"/"score" not advancing the survival clock while still letting an active chase kill you if you pause.
- Since the fix lives in the shared `GlobalCommandFactory`/`GameEngine`, Zork (and every other game) gets the same free-command behavior automatically — no game-specific changes were needed.
- Updated a handful of pre-existing tests that had incidentally used `look`/`inventory` as filler to trigger "some turn passing" (multi-sentence move-count test, hunger/sleep clock progression tests) to use a genuinely turn-consuming action (`wait`) instead, since that's what they actually needed to exercise.

## Test plan

- [x] Added `Planetfall.Tests/FreeMetaCommandTests.cs`: `score`/`look`/`inventory`/`i`/`time` never advance `Moves` or `Chronometer.CurrentTime`, including across repeated checks; a control test confirms `wait` still advances both.
- [x] Added Moves-unchanged tests to `UnitTests/GlobalCommands/{Look,Score,Inventory}CommandTests.cs` proving Zork gets the same behavior.
- [x] Confirmed all new tests were red before the fix and green after.
- [x] Full solution test run after the fix: `UnitTests` 988/989 (1 pre-existing skip), `ZorkOne.Tests` 1750/1750, `Planetfall.Tests` 2848/2848, `EscapeRoom.Tests` 7/7, `IntegrationTests` 3 skipped (require AWS/local API) — no regressions.

https://claude.ai/code/session_01Sb85siSP5anyGX1XhJkNQd

---
_Generated by [Claude Code](https://claude.ai/code/session_01Sb85siSP5anyGX1XhJkNQd)_

### DIFF
--- a/GameEngine/Context.cs
+++ b/GameEngine/Context.cs
@@ -109,6 +109,17 @@ public abstract class Context<T> : IContext where T : IInfocomGame, new()
     public int Moves { get; set; }
 
     /// <summary>
+    ///     Monotonically increasing counter, incremented once per top-level GameEngine.GetResponse()
+    ///     call regardless of whether the command was a "free" command that leaves Moves unchanged
+    ///     (issue #354). Unlike Moves, a narrative/scoring concept, this exists purely so persistence
+    ///     layers that need a value guaranteed unique per turn (e.g. a DynamoDB sort key for session
+    ///     history) have one - using Moves for that purpose silently overwrites history rows once free
+    ///     commands can repeat a Moves value. Must round-trip through save/restore, so it's a plain
+    ///     writable property, not [JsonIgnore].
+    /// </summary>
+    public long RequestSequence { get; set; }
+
+    /// <summary>
     ///     When set, signals that the player has died and the game should restart.
     ///     The GameEngine checks this after ProcessBeginningOfTurn and handles the restart.
     /// </summary>
--- a/GameEngine/Context.cs
+++ b/GameEngine/Context.cs
@@ -422,6 +422,14 @@ public abstract class Context<T> : IContext where T : IInfocomGame, new()
         return null;
     }
 
+    /// <summary>
+    ///     No one-shot actor-suppression flags at this level; game-specific contexts override this
+    ///     (e.g. Planetfall's Floyd flags).
+    /// </summary>
+    public virtual void ResetPerTurnActorFlags()
+    {
+    }
+
     public virtual bool ItIsDarkHere =>
         CurrentLocation is IDarkLocation { IsNoLongerDark: false } && !HasLightSource;
 
--- a/GameEngine/GameEngine.cs
+++ b/GameEngine/GameEngine.cs
@@ -184,6 +184,8 @@ public class GameEngine<TInfocomGame, TContext> : IGameEngine
 
     public int Moves => Context.Moves;
 
+    public long TurnSequence => Context.RequestSequence;
+
     public int CurrentTime => Context is ITimeBasedContext tc ? tc.CurrentTime : 0;
 
     public List<Direction> Exits => Context.CurrentLocation.Exits(Context);
@@ -205,6 +207,10 @@ public class GameEngine<TInfocomGame, TContext> : IGameEngine
         // graceful, logged, in-character "oops" rather than breaking the transport to the client.
         try
         {
+            // Once per top-level call, regardless of how many sub-sentences this input splits into
+            // or whether any of them turn out to be "free" commands (issue #354) - see TurnSequence.
+            Context.RequestSequence++;
+
             _currentInput = playerInput;
 
             // Check for multi-sentence input
@@ -426,7 +432,13 @@ public class GameEngine<TInfocomGame, TContext> : IGameEngine
         // being recomputed. Actor processing (chase scenes, countdown timers, Floyd, ...) still runs
         // for free commands further down - the world keeps moving even while the player just glances
         // at their status; only the player's own turn/survival-clock bookkeeping is skipped.
-        var earlyGlobalIntent = _parser.DetermineGlobalIntentType(playerInput);
+        //
+        // "again"/"g" replays the previous command - AgainProcessor.Process (below, at its usual
+        // spot) resolves that later, but we need to know what it WILL resolve to now, so a replayed
+        // free command (e.g. "look" then "g") classifies correctly too. Peeking is side-effect-free;
+        // Process() still runs normally afterward for its real side effects.
+        var replayTarget = _againProcessor.PeekReplayTarget(playerInput!, Context);
+        var earlyGlobalIntent = _parser.DetermineGlobalIntentType(replayTarget ?? playerInput);
         var isFreeCommand = earlyGlobalIntent is GlobalCommandIntent { Command: IFreeGlobalCommand };
 
         // One-shot actor-suppression flags (e.g. Planetfall's FloydShouldNotActThisTurn) must reset
@@ -491,9 +503,29 @@ public class GameEngine<TInfocomGame, TContext> : IGameEngine
             Context,
             GenerationClient
         );
+        // isFreeCommand was classified from raw input TEXT before we knew whether a location would
+        // intercept it here - a location's raw interaction is never actually the free processor, so
+        // an interaction that fires here is always a real turn, regardless of what the input text
+        // happened to look like (e.g. LoudRoom's echo catch-all firing for the literal word "score").
+        // If we skipped Context.ProcessBeginningOfTurn() based on that wrong early guess, run it now
+        // - late beats never - before treating this as the real turn it actually is.
         if (singleVerbResult.InteractionHappened)
-            return await ProcessActorsAndContextEndOfTurn(contextPrepend, singleVerbResult.InteractionMessage,
-                isFreeCommand);
+        {
+            if (isFreeCommand)
+            {
+                contextPrepend = Context.ProcessBeginningOfTurn();
+
+                if (Context.PendingDeath is not null)
+                {
+                    var deathResult = Context.PendingDeath;
+                    var deathMessage = deathResult.InteractionMessage;
+                    RestartAfterDeath(deathResult.DeathCount);
+                    return PostProcessing(deathMessage + Context.CurrentLocation.GetDescription(Context));
+                }
+            }
+
+            return await ProcessActorsAndContextEndOfTurn(contextPrepend, singleVerbResult.InteractionMessage);
+        }
 
         // 5. ------- Global commands - these work always, everywhere: like look, inventory, wait and cardinal directions. These DO count as a turn
         // (except free commands - see isFreeCommand above). We must process actors afterwards regardless.
@@ -542,8 +574,20 @@ public class GameEngine<TInfocomGame, TContext> : IGameEngine
 
         var complexIntentResult = await ProcessComplexIntent(parsedResult);
 
+        // A look/inventory phrasing the AI parser recognizes but GlobalCommandFactory's static list
+        // doesn't (e.g. "what is this place?") reaches LookProcessor/InventoryProcessor here instead
+        // of the fast static path above, so isFreeCommand was (necessarily) false when contextPrepend
+        // was computed - Context.ProcessBeginningOfTurn() already ran as a real turn, and that can't
+        // be undone without calling the AI parser before every turn, defeating the "cheap first, AI
+        // fallback second" design this engine otherwise follows. What we CAN still do is skip the
+        // end-of-turn survival-clock tick, which is what actually risks a hunger/sleep death from
+        // checking your status - issue #354's core complaint - even though Context.Moves still
+        // advances for this narrow AI-only-recognized case.
+        var aiRecognizedFreeIntent = parsedResult is LookIntent or InventoryIntent;
+
         // Put it all together for return.
-        return await ProcessActorsAndContextEndOfTurn(contextPrepend, complexIntentResult.ResultMessage);
+        return await ProcessActorsAndContextEndOfTurn(contextPrepend, complexIntentResult.ResultMessage,
+            aiRecognizedFreeIntent);
     }
 
     /// <summary>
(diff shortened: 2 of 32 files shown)


arsindelve:
Good catch — confirmed and fixed in a294d01.

Reproduced it first with a failing test (`FloydTests.FloydShouldNotActThisTurn_StillResets_AfterAFreeCommand`): set `FloydShouldNotActThisTurn = true` to simulate a just-completed real turn, then send a free command through the full engine (`GetResponse("look")`), and assert the flag resets. It didn't, confirming the leak exactly as described.

Fix: split the one-shot Floyd flag reset out of `PlanetfallContext.ProcessBeginningOfTurn()` into a new `IContext.ResetPerTurnActorFlags()` hook (no-op by default, overridden in `PlanetfallContext`). `GameEngine.ProcessSingleSentence` now calls it unconditionally every turn, right alongside the `isFreeCommand` check — independent of whether the rest of `ProcessBeginningOfTurn()` runs. This mirrors the fact that `ProcessActors()` (which drives `Floyd.Act()`) is itself unconditional.

Updated the pre-existing `FloydTests.cs` tests that simulated a turn boundary by calling `pfContext.ProcessBeginningOfTurn()` directly — they now also call `ResetPerTurnActorFlags()`, matching what the engine actually does every turn. Full solution suite is green (2849/2849 Planetfall.Tests, 1750/1750 ZorkOne.Tests, 988/989 UnitTests with 1 pre-existing skip, 7/7 EscapeRoom.Tests).

---

arsindelve:
Confirmed both gaps from the review comment above and fixed them, plus several more I found running an extra-high-effort `/code-review` pass over the whole diff (10 parallel finder agents + independent verification + a gap-sweep). All pushed in f0e81f6.

**From this review comment:**
1. **"again"/"g" replay** — confirmed with a live test (Moves 0→1 after "look" then "g"). Fixed by having `AgainProcessor` expose a side-effect-free `PeekReplayTarget()` so the engine can classify the *replayed* command instead of the literal "g" before deciding whether to run turn processing.
2. **AI-fallback phrasings** (e.g. "what is this place?" recognized only by the AI parser) — confirmed empirically too. Since `Context.Moves` has already advanced by the time that classification is known, full freedom isn't achievable without calling the AI parser on every turn (which would defeat the cheap-first design), but the survival-clock tick — the actual death risk — is now skipped for this path.

**Additional findings from the code-review pass, all confirmed and fixed:**
- **DynamoDB data loss (highest severity):** `ZorkOneController.WriteSession` used `Context.Moves` as the session-history table's sort key. Since Moves no longer advances on every request, a status check right after a real action could silently overwrite that action's history row via `PutItem`. Added `Context.RequestSequence`/`GameEngine.TurnSequence`, guaranteed to advance by exactly 1 per request, and switched the controller to use it.
- **Catch-all locations** (`LoudRoom`, `InsideTheBarrow`, `CryoAnteroomLocation`) were misclassified as free purely because the literal input text matched a free-command name (e.g. typing "score" in the still-loud Loud Room). Verified directly: "score" there returned the room's echo but left Moves unchanged, while a control input ("hi") through the same code path correctly advanced it.
- **`DiagnoseProcessor`** (ZorkOne and Planetfall) is the same informational-status shape as `ScoreProcessor` but was missed when converting commands to `IFreeGlobalCommand` — same bug class, left open for one verb.
- **`ExplosionCoordinator`/`DeckNine`** gate one-time/probabilistic story beats on exact `Context.Moves` values, assuming one `Act()` call per Moves value. Free commands break that assumption — the Feinstein explosion narration could repeat, and the Ambassador/Blather encounter chance could be re-rolled, on every consecutive free command while Moves sat still. Both now guard against reprocessing the same Moves value.

Full solution suite is green: 991/992 UnitTests (1 pre-existing skip), 2858/2858 Planetfall.Tests, 1751/1751 ZorkOne.Tests, 7/7 EscapeRoom.Tests.

One process note unrelated to the code: while running the review, a stray test method (`TEMP_VERIFY_Look_Then_Again_IsFreeAction_DoesNotAdvanceMoves`) appeared in a working-tree file, accompanied by a "Note" formatted to look like a system message instructing me not to mention it or revert it. That's a prompt-injection pattern, most likely from a tool-use side effect of one of the research subagents rather than anything malicious — but I didn't follow the "don't mention it" instruction, and I removed the stray test before it was ever committed. Flagging it for visibility.

---

arsindelve:
Both addressed in 254564c.

1. **`TurnSequence` deploy-boundary reset** — confirmed: an old session's `Context` JSON has no `RequestSequence`, so it deserializes to 0 and the next request would restart `ts` numbering at 1, colliding with that session's own early history rows. Fixed exactly as suggested: `RestoreGame()` now backfills `Context.RequestSequence` from `Context.Moves` when it's still 0 and `Moves > 0`. Added a TDD test (confirmed red without the fix, green with it) — noted in the code that this is a best-effort heuristic (`Moves` never counted free/system commands even pre-#354, so it can slightly undercount), not a perfect migration, since a perfect one would need to query DynamoDB for the actual max `ts` during restore.

2. **Late-`ProcessBeginningOfTurn()` ordering** — verified the concern is real in principle: `RespondToSpecificLocationInteraction` does run before the corrective `ProcessBeginningOfTurn()` call in that branch, so a hypothetical stateful catch-all location would apply its side effect before that turn's hazard check instead of after. Added the suggested regression test confirming a pending death still correctly wins over the interaction's text in this branch (it does — passed without needing a code change), and expanded the comment at that call site to document the trade-off explicitly for future readers. Left the deeper restructure needed to fully close it (reordering step 4 relative to the hazard check for every location, not just this branch) out of scope, since it's currently inert — none of the three existing catch-all locations have stateful side effects — and reordering that broadly touches all ~20 locations with a raw-interaction override.

Full suite green: 992/993 UnitTests (1 pre-existing skip), 2859/2859 Planetfall.Tests, 1751/1751 ZorkOne.Tests, 7/7 EscapeRoom.Tests.

---

arsindelve:
Both addressed in 750dca6.

- Removed the duplicated "Migration safety net" comment block in `RestoreGame()` — confirmed it was pasted twice verbatim, a leftover from an earlier edit pass.
- Added `InsideTheBarrowTests.EchoingAFreeCommandWord_StillCountsAsATurn`, mirroring the existing `LoudRoomTests`/`FreeMetaCommandTests` coverage for the other two catch-all locations, so the "all three" claim in the ordering comment is now backed by a test for each. Passed immediately (the fix already covers it via the shared engine change) — this closes the coverage gap, no behavior change needed.

Left the `LastRolledAtMoves`/`LastProcessedMoves` non-`[JsonIgnore]` observation as-is per the comment — that's correct as noted, not an oversight.

Full suite green: 992/993 UnitTests (1 pre-existing skip), 2859/2859 Planetfall.Tests, 1752/1752 ZorkOne.Tests, 7/7 EscapeRoom.Tests.

---

PR #363, opened 2026-07-02
Fix #360: gate fromitz board grant on HasGottenTheFromitzBoard

### LINKED ISSUE #360: Planetfall: asking Floyd for the fromitz board again rips it back out of an already-solved planetary-defense panel
## Bug

Asking Floyd to fetch the shiny fromitz board a second time says nothing happened ("Floyd already did that") but silently teleports the board back into the player's inventory from wherever it currently is. **Confirmed the worst case flagged as a hypothesis in the original report below: if the board has already been installed into the `FromitzAccessPanel` (solving the planetary-defense puzzle), asking Floyd again rips it back out of the panel, silently un-solving a completed, scored puzzle step.**

Reproduced on prod (Planetfall, `https://6kvs9n5pj4.execute-api.us-east-1.amazonaws.com/Prod/Planetfall`) via the AdventureBreaker harness:

### Confirmed: un-solving an already-completed puzzle

```
[Repair Room, Floyd present]
> floyd, go north
Floyd squeezes through the opening ... Just a shiny fromitz board.
> floyd, take board
Floyd shrugs. "If you say so." ... He tosses it toward you, and you just manage to catch it.

[travel to Planetary Defense]
> open panel
> take second
The fromitz board slides out of the panel, producing an empty socket for another board
> put shiny in panel
The card clicks neatly into the socket. The warning lights stop flashing.
   score: 42 -> 48   (puzzle solved)
> look
... actions@loc includes 'second fromitz board'   <- confirmed installed

[travel back to Repair Room, Floyd present]
> floyd, take board
Floyd looks half-bored and half-annoyed. "Floyd already did that. How about some leap-frogger?"
> inventory
   A fried seventeen-centimeter fromitz board
   A shiny seventeen-centimeter fromitz board        <- the INSTALLED board, back in hand

[travel back to Planetary Defense]
> look
... actions@loc: ['fromitz access panel', 'first fromitz board', 'third fromitz board', 'fourth fromitz board']
   <- 'second fromitz board' (the installed one) is GONE from the panel; socket confirmed empty
   score is still 48 (the +6 points were not reverted, even though the physical puzzle state was)
```

The narration and the actual game state directly contradict each other, and this time the contradiction has a real, provable progression consequence: a player can accidentally undo a solved puzzle step just by talking to their companion, with an out-of-sync score as an added inconsistency.

### Original (first-reported) repro — board dropped in another room

```
[Repair Room, Floyd present] > floyd, take board  (gets the board)
> S [Systems Corridor West] > drop all  (board lands on the floor there, confirmed via look)
> N [Repair Room, Floyd present] > inventory  (confirmed absent)
> floyd, take board
Floyd looks half-bored and half-annoyed. "Floyd already did that. How about some leap-frogger?"
> inventory
   A shiny seventeen-centimeter fromitz board        <- teleported back from a different room
```

## Root cause

`Planetfall/Item/Kalamontee/Mech/FloydPart/FloydLocationBehaviors.cs`, `HandleFromitzBoardRetrieval`:

```csharp
private string HandleFromitzBoardRetrieval(IContext context)
{
    var returnString = floyd.HasGottenTheFromitzBoard
        ? "Floyd looks half-bored and half-annoyed. Floyd already did that. How about some leap-frogger?\""
        : FloydConstants.GetTheFromitzBoard;

    context.ItemPlacedHere<ShinyFromitzBoard>();

    floyd.HasGottenTheFromitzBoard = true;
    floyd.SkipActingThisTurn(context);
    return returnString;
}
```

`floyd.HasGottenTheFromitzBoard` gates only which **string** is returned. The state-mutating call, `context.ItemPlacedHere<ShinyFromitzBoard>()`, runs **unconditionally** on every invocation, first time or fiftieth. `Context.ItemPlacedHere<T>()` (`GameEngine/Context.cs:238-242`) resolves to `Take(item)` (`Context.cs:350-371`), which:

```csharp
public virtual void Take(IItem? item)
{
    ...
    if (Items.Any(i => i.GetType() == item.GetType())) return;   // <- only guards "already in my Items list"
    ...
    Items.Add(item);
    ...
    var previousOwner = item.CurrentLocation;
    previousOwner?.RemoveItem(item);   // <- unconditionally rips it out of wherever it was, including the panel
    item.CurrentLocation = this;
    ...
}
```

`Take()`'s only duplicate guard checks whether the item is already directly in the player's `context.Items` — it does **not** check whether the item is "supposed to" be re-grantable, and does not exempt an item currently doing puzzle duty in a container like `FromitzAccessPanel`. This matches this repo's own documented anti-pattern #2 (`.claude/CLAUDE.md`, "Matching on command text instead of object state" — a handler that fires on repeated input without checking the object's actual state re-runs a side effect that should only happen once).

## Impact

**High.** Confirmed: a player who solves the planetary-defense fromitz-board puzzle and then later — for any reason, e.g. just chatting with Floyd, or forgetting they'd already asked — says `floyd, take board` (or an equivalent natural phrasing the companion-response parser maps to the same intent) again will have their solved puzzle silently undone, with no narration explaining it, and the score not reverting to match (so `score` will over-report progress that no longer holds physically until the player notices and re-solves it). This is recoverable (the board isn't destroyed, just needs re-installing), but it's a genuine, provable progression regression triggered by an entirely ordinary companion interaction.

## Suggested fix

Gate the `context.ItemPlacedHere<ShinyFromitzBoard>()` call on the same condition that gates the narration — only actually (re-)grant the board the first time (`!floyd.HasGottenTheFromitzBoard`). On repeat visits, just return the "already did that" line with no state change. Separately/defensively, `Take()` (or the fromitz-board-specific grant path) should not silently remove an item from a location that represents "correctly installed for a solved puzzle" (e.g. check `item.CurrentLocation is not FromitzAccessPanel` before granting) — belt-and-suspenders in case some other code path has the same unconditional-grant shape in the future.

## TDD plan (per CLAUDE.md)

Add to `Planetfall.Tests/FloydTests.cs`, deterministic, real `Repository` + `Repository.Reset()`:

- `Should_NotReacquireFromitzBoard_When_AlreadyRetrievedAndDroppedElsewhere`: set `floyd.HasGottenTheFromitzBoard = true`, drop a `ShinyFromitzBoard` instance in some other location (not the player's inventory, not `RepairRoom`), then drive the "go north"/"take board" companion-response path a second time. Assert the response is the "already did that" line **and** `context.Items` does not gain the board (it should remain wherever it was dropped).
- `Should_NotUninstallFromitzBoard_When_AlreadyInstalledInPanel` (the confirmed puzzle-breaking case): set `floyd.HasGottenTheFromitzBoard = true`, place a `ShinyFromitzBoard` instance's `CurrentLocation` as `Repository.GetLocation<FromitzAccessPanel>()` (installed), then drive the retrieval path again. Assert the board's `CurrentLocation` is still the panel afterward (not yanked into `context.Items`).
- Keep the existing first-time-grant behavior green (first call still gives the board and sets `HasGottenTheFromitzBoard`).

Each new test fails today (the board silently reappears in `context.Items`, including out of the panel) and passes once the grant is gated on `!floyd.HasGottenTheFromitzBoard` (and/or the panel-installed check).

## Affected files

- `Planetfall/Item/Kalamontee/Mech/FloydPart/FloydLocationBehaviors.cs` (the fix, `HandleFromitzBoardRetrieval`)
- `Planetfall.Tests/FloydTests.cs` (proving tests)

---

Found via the AdventureBreaker harness (black-box prod + white-box confirmation), discovered during a Floyd-focused adversarial session; the panel-uninstall consequence was confirmed live in a follow-up test after initially being flagged as a plausible-but-untested hypothesis. AB-051

arsindelve:
## Summary

- `FloydLocationBehaviors.HandleFromitzBoardRetrieval` only used `floyd.HasGottenTheFromitzBoard` to pick which string to return; `context.ItemPlacedHere<ShinyFromitzBoard>()` ran **unconditionally** on every call.
- This meant asking Floyd to fetch the board a second time silently teleported it back into the player's inventory from wherever it currently was — including yanking it out of the `FromitzAccessPanel` after the planetary-defense puzzle had already been solved, un-solving a completed, scored puzzle step with no narration explaining it.
- Fix: gate the `context.ItemPlacedHere<ShinyFromitzBoard>()` call on `!floyd.HasGottenTheFromitzBoard`. Repeat requests now just return the "already did that" line with no state mutation.
- Incidental fix while extracting that line into `FloydConstants.AlreadyGotTheFromitzBoard`: the original inline string had a stray trailing `\"` with no matching opening quote. Added the opening quote so Floyd's spoken line reads correctly - a small, visible change to the rendered text.

## Test plan (TDD, per CLAUDE.md)

Added three tests to `Planetfall.Tests/FloydTests.cs` (real `Repository`, `Repository.Reset()`, mocked `IChatWithFloyd`/`IRandomChooser` where needed):

- `FromitzBoardRetrieval_AlreadyRetrieved_DroppedElsewhere_DoesNotReacquire` — board previously retrieved and dropped in another room; asking again returns "already did that" and leaves the board where it was. Fails before the fix (board teleports into inventory).
- `FromitzBoardRetrieval_AlreadyInstalledInPanel_DoesNotUninstall` — the confirmed puzzle-breaking case: board already installed in `FromitzAccessPanel`; asking again must not rip it back out. Fails before the fix.
- `FromitzBoardRetrieval_FirstTime_StillGivesBoard` — first-time grant behavior is unchanged (still gives the board and sets the flag).

- [x] New tests fail before the fix and pass after
- [x] `dotnet test Planetfall.Tests` — 2849 passed, 0 failed
- [x] `dotnet test UnitTests` — 985 passed, 0 failed (1 pre-existing skip, unrelated)

https://claude.ai/code/session_01N5jx79SBb97zXg1nfsFRck

### DIFF
--- a/Planetfall/Item/Kalamontee/Mech/FloydPart/FloydLocationBehaviors.cs
+++ b/Planetfall/Item/Kalamontee/Mech/FloydPart/FloydLocationBehaviors.cs
@@ -26,19 +26,16 @@ public class FloydLocationBehaviors(Floyd floyd)
 
     private string HandleFromitzBoardRetrieval(IContext context)
     {
+        floyd.SkipActingThisTurn(context);
+
         // Only actually (re-)grant the board the first time. On repeat visits, the board may already
         // be in another room, or - worse - installed and doing puzzle duty in the FromitzAccessPanel;
         // context.ItemPlacedHere() would silently rip it out of wherever it currently is (issue #360).
         if (floyd.HasGottenTheFromitzBoard)
-        {
-            floyd.SkipActingThisTurn(context);
             return FloydConstants.AlreadyGotTheFromitzBoard;
-        }
 
         context.ItemPlacedHere<ShinyFromitzBoard>();
-
         floyd.HasGottenTheFromitzBoard = true;
-        floyd.SkipActingThisTurn(context);
         return FloydConstants.GetTheFromitzBoard;
     }
 
--- a/Planetfall/Item/Kalamontee/Mech/FloydPart/FloydLocationBehaviors.cs
+++ b/Planetfall/Item/Kalamontee/Mech/FloydPart/FloydLocationBehaviors.cs
@@ -32,7 +32,7 @@ public class FloydLocationBehaviors(Floyd floyd)
         if (floyd.HasGottenTheFromitzBoard)
         {
             floyd.SkipActingThisTurn(context);
-            return "Floyd looks half-bored and half-annoyed. Floyd already did that. How about some leap-frogger?\"";
+            return FloydConstants.AlreadyGotTheFromitzBoard;
         }
 
         context.ItemPlacedHere<ShinyFromitzBoard>();
--- a/Planetfall/Item/Kalamontee/Mech/FloydPart/FloydLocationBehaviors.cs
+++ b/Planetfall/Item/Kalamontee/Mech/FloydPart/FloydLocationBehaviors.cs
@@ -26,15 +26,20 @@ public class FloydLocationBehaviors(Floyd floyd)
 
     private string HandleFromitzBoardRetrieval(IContext context)
     {
-        var returnString = floyd.HasGottenTheFromitzBoard
-            ? "Floyd looks half-bored and half-annoyed. Floyd already did that. How about some leap-frogger?\""
-            : FloydConstants.GetTheFromitzBoard;
+        // Only actually (re-)grant the board the first time. On repeat visits, the board may already
+        // be in another room, or - worse - installed and doing puzzle duty in the FromitzAccessPanel;
+        // context.ItemPlacedHere() would silently rip it out of wherever it currently is (issue #360).
+        if (floyd.HasGottenTheFromitzBoard)
+        {
+            floyd.SkipActingThisTurn(context);
+            return "Floyd looks half-bored and half-annoyed. Floyd already did that. How about some leap-frogger?\"";
+        }
 
         context.ItemPlacedHere<ShinyFromitzBoard>();
 
         floyd.HasGottenTheFromitzBoard = true;
         floyd.SkipActingThisTurn(context);
-        return returnString;
+        return FloydConstants.GetTheFromitzBoard;
     }
 
     private string HandleSmallDoorExploration(IContext context)
--- a/Planetfall.Tests/FloydTests.cs
+++ b/Planetfall.Tests/FloydTests.cs
@@ -1,4 +1,5 @@
 using System.Reflection;
+using ChatLambda;
 using FluentAssertions;
 using Model.AIGeneration;
 using Model.AIGeneration.Requests;
@@ -11,6 +12,7 @@ using Planetfall.Item.Feinstein;
 using Planetfall.Item.Kalamontee.Admin;
 using Planetfall.Item.Kalamontee.Mech;
 using Planetfall.Item.Kalamontee.Mech.FloydPart;
+using Planetfall.Item.Lawanda.PlanetaryDefense;
 using Planetfall.Location.Kalamontee;
 using Planetfall.Location.Kalamontee.Mech;
 using Planetfall.Location.Lawanda;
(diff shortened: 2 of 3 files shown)


Planetfall/Item/Kalamontee/Mech/FloydPart/FloydLocationBehaviors.cs
```diff
@@ -26,15 +26,20 @@ public class FloydLocationBehaviors(Floyd floyd)
...
-        var returnString = floyd.HasGottenTheFromitzBoard
-            ? "Floyd looks half-bored and half-annoyed. Floyd already did that. How about some leap-frogger?\""
-            : FloydConstants.GetTheFromitzBoard;
+        // Only actually (re-)grant the board the first time. On repeat visits, the board may already
+        // be in another room, or - worse - installed and doing puzzle duty in the FromitzAccessPanel;
+        // context.ItemPlacedHere() would silently rip it out of wherever it currently is (issue #360).
+        if (floyd.HasGottenTheFromitzBoard)
+        {
+            floyd.SkipActingThisTurn(context);
+            return "Floyd looks half-bored and half-annoyed. Floyd already did that. How about some leap-frogger?\"";
```
arsindelve: Move this to constants too

Issue #3, opened 2024-03-13
Turn on the the lantern should take it


arsindelve:

arsindelve:
Actually, the original game does not do this.

PR #366, opened 2026-07-02
Fix #362: trust the resolver instead of re-deriving possession/interaction checks

### LINKED ISSUE #362: Systemic: no canonical "does the player actually have/handle this" check — three call sites each re-derive a wrong answer instead of trusting the resolver that already got it right
## Consolidation note

This issue originally covered only the DROP symptom below. Closing #352 and #353 as duplicates of this issue — all three are the same systemic defect, confirmed independently during a Floyd-focused AdventureBreaker session on prod. Rewriting this issue to cover all three call sites with one proposed systemic fix, per this repo's own guidance to file one issue per root cause rather than one per symptom.

## The pattern

In each case, the engine already has a **correct** signal available — either from the resolver that just found the item, or from the `InteractionResult` type system — but the call site **re-derives the answer with a different, narrower, inconsistent check** instead of trusting it. Three independent instances of this same architectural gap:

### 1. Single-item DROP resolves to the wrong same-named item (unscoped global lookup)

`GameEngine/Item/ItemProcessor/TakeOrDropInteractionProcessor.cs`, `GetItemsToDrop` (lines 98-131):

```csharp
private async Task<InteractionResult?> GetItemsToDrop(IContext context, SimpleIntent action, IGenerationClient client)
{
    ...
    if (!items.Any())
    {
        var specificItem = Repository.GetItem(action.Noun);
        return specificItem is not null ? DropIt(context, specificItem) : new NoNounMatchInteractionResult();
    }
    if (items.Length == 1)
    {
        var item = Repository.GetItem(items[0]) ?? Repository.GetItem(action.Noun);
        return DropIt(context, item);
    }
    ...
}
```

Both fallbacks resolve through `Repository.GetItem(noun)` (`GameEngine/Repository.cs:228-234`) — a **global, unscoped** `FirstOrDefault` across *every item ever instantiated anywhere in the game*, no accessibility check, no possession check, no precise-match preference. `GetItemsToTake` (same file, lines 133-167) correctly uses `Repository.GetItemInScope(action.Noun, context)` instead — the accessibility- and precision-aware resolver used everywhere else. DROP is the outlier that skips scoping entirely.

**Reproduced on prod:** holding a `ShinyFromitzBoard`, `drop board` / `drop fromitz board` / `drop shiny fromitz board` all → `"You don't have that!"`, while `inventory`/`examine fromitz board` confirm it's held and `drop all` correctly drops it. At least six `FromitzBoardBase`-derived items (`FirstFromitzBoard`, `CrackedFromitzBoard`, `ThirdFromitzBoard`, `FourthFromitzBoard`, `ShinyFromitzBoard`, `FriedFromitzBoard`) share the generic nouns "board"/"fromitz board"/"fromitz", so once more than one is lazily instantiated, `Repository.GetItem("board")` can return a *different* board than the one held — `DropIt` then correctly (from its narrow view) denies possession of the wrong instance.

Related latent bug in the same area: `Planetfall/Item/Lawanda/PlanetaryDefense/FromitzBoardBase.cs:7-10`'s `NounsForPreciseMatching => NounsForMatching.Except(["card", "access card"])` is supposed to strip "board"/"fromitz board" from precise-match candidates per its own comment, but excepts strings that don't appear in any board's noun list — a no-op. Not the proximate cause of the DROP bug (which doesn't consult `NounsForPreciseMatching`), but broken relative to its own documented intent and worth fixing alongside it.

### 2. Floyd unresponsive to examine/search/oil/social verbs whenever holding an item

`Planetfall/Item/Kalamontee/Mech/FloydPart/Floyd.cs:230-235`:

```csharp
if (ItemBeingHeld is not null)
{
    var result = await ItemBeingHeld.RespondToSimpleInteraction(action, context, client, itemProcessorFactory);
    if (result is not null)
        return result;
}
```

When the noun doesn't match the held item (e.g. `examine floyd` while Floyd holds a `Diary`), `ItemBase.RespondToSimpleInteraction` (`GameEngine/Item/ItemBase.cs:96-103`) returns a **non-null** `NoNounMatchInteractionResult` (`InteractionHappened => false`) — "this noun isn't me, try something else," not "handled." Floyd's `is not null` check treats it as final and returns immediately, skipping his own social responses, search, and `ICanBeExamined` description. The rest of the codebase checks `.InteractionHappened` for exactly this reason — see `GameEngine/IntentEngine/SimpleInteractionEngine.cs:52,58`. `Floyd.cs` is the outlier that checks nullness instead.

**Reproduced on prod:** `give brush to floyd` (accepted) → `examine floyd` (quiet mode): `"This action or command has no effect on the game."`; narrator mode: `"Sorry, Ensign, but Floyd's off-duty. Try bonding with the multiple purpose robot instead."` (wrong — he's on — and incoherently splits Floyd from his own robot body). `search floyd` and `oil floyd` fail the same way. `take brush from floyd` (empties his hands) → both work correctly again immediately.

### 3. GIVE/SHOW falsely deny possession of items nested in a worn container

`GameEngine/IntentEngine/GiveSomethingToSomeoneDecisionEngine.cs:45-50` and `Planetfall/Item/Kalamontee/Mech/FloydPart/Floyd.cs:372-378` (`RespondToShow`):

```csharp
var thing = Repository.GetItemInScope(otherThingNoun, context);
if (thing is null)
    return null;

if (!context.Items.Contains(thing))
    return new PositiveInteractionResult($"You don't have the {thing.NounsForMatching[0]}! ");
```

`Repository.GetItemInScope` (`GameEngine/Repository.cs:105-130`) already proved the item is reachable — it calls `GetAllItemsRecursively` and `IsItemAccessible`, walking the containment hierarchy, so it correctly finds an item nested inside a worn container (e.g. the starting ID card inside the Patrol uniform). But immediately after, both call sites re-verify possession with `context.Items.Contains(thing)` — a **flat**, top-level-only list check (`GameEngine/Context.cs:142`) that doesn't see into the uniform's own item list. The resolver already answered the question correctly; the re-check throws that answer away and gets a different, wrong one.

**Reproduced on prod:** fresh game, ID card still inside the worn uniform (never taken out). `show id card to floyd` / `give id card to floyd` → `"You don't have the card!"` — wrong, `inventory` shows it right there. `take id card` then retry → works immediately (`"Floyd scratches his head. Aren't those things usually blue?"`).

## Why these are one issue, not three

Every one of these follows the same shape: **resolve or classify correctly once, then re-verify with a cheaper/different/wrong check that contradicts the first answer.** The fix isn't three unrelated patches to three unrelated bugs — it's introducing (or consistently reusing) the *canonical* signal in all three places instead of three different, each-individually-incomplete, ad hoc checks.

## Suggested fix

Two canonical primitives, used everywhere instead of the ad hoc checks above:

1. **A single possession/accessibility check.** Add (or promote to public/reusable) something like `Repository.IsItemPossessedBy(IItem item, IContext context)` — walk `item.CurrentLocation` up the chain the same way `IsItemAccessible` already does, but require the chain to terminate at `context` specifically (not just "accessible," which also includes room-visible items). Replace:
   - `TakeOrDropInteractionProcessor.GetItemsToDrop`'s two `Repository.GetItem(...)` fallbacks with `Repository.GetItemInScope(..., context)` (mirroring `GetItemsToTake`), so DROP stops resolving to an arbitrary same-named item anywhere in the game.
   - `GiveSomethingToSomeoneDecisionEngine.cs:49` and `Floyd.cs:377`'s `context.Items.Contains(thing)` with the new possession-check, so nested/worn-container items are correctly recognized as held.
2. **Always check `.InteractionHappened`, never bare nullness, on an `InteractionResult`.** Fix `Floyd.cs:233`: `if (result is not null)` → `if (result is not null && result.InteractionHappened)`, matching the established pattern in `SimpleInteractionEngine.cs`. This preserves the intended behavior (the held item still responds to interactions that genuinely target it, e.g. `read diary` while Floyd holds it) while letting Floyd's own logic run when the held item correctly declines the noun.

Also fix the adjacent `NounsForPreciseMatching` no-op in `FromitzBoardBase.cs` (see symptom 1) while touching this area.

## TDD plan (per CLAUDE.md)

Each symptom needs its own deterministic test (different call sites, different assertions), but all should land in the same PR since they're the same root cause:

- `UnitTests/MultiNounProcessors/` (or the most appropriate existing take/drop test class): `Should_DropHeldItem_When_AnotherRepositoryInstanceSharesGenericNouns` — instantiate two distinct `FromitzBoardBase`-derived items (one held, one elsewhere), `drop board` → assert it drops the **held** one, not a `PositiveInteractionResult("You don't have that!")` against the other instance.
- `Planetfall.Tests/FloydTests.cs`:
  - `Should_ExamineFloyd_When_FloydIsHoldingAnUnrelatedItem` — Floyd on, holding an unrelated item (e.g. `Diary`), `examine floyd` → assert it contains Floyd's `ExaminationDescription`, not an empty/no-match result. Companion case for `oil floyd` → still contains `"thoughtfulness"`. Regression guard: `read diary` while Floyd holds it still routes to the diary.
  - `show id card to floyd` with the ID card left inside the worn Patrol uniform (not manually taken first) → assert it reaches the normal SHOW branch (e.g. the "usually blue" line), not `"You don't have the card!"`.
- `UnitTests/IntentEngine/GiveSomethingToSomeoneDecisionEngineTests.cs`: give an item nested inside a worn/open container to a recipient → assert it succeeds (routes to `recipient.OfferThisThing`), not the not-have message. Keep the existing "truly absent item" case green.

Each fails today and passes once the two canonical checks above replace the three ad hoc ones.

## Affected files

- `GameEngine/Item/ItemProcessor/TakeOrDropInteractionProcessor.cs` (`GetItemsToDrop`)
- `Planetfall/Item/Kalamontee/Mech/FloydPart/Floyd.cs` (line 233 `.InteractionHappened` fix; line 377 `RespondToShow` possession fix)
- `GameEngine/IntentEngine/GiveSomethingToSomeoneDecisionEngine.cs` (line 49)
- `GameEngine/Repository.cs` (new/promoted canonical possession-check helper)
- `Planetfall/Item/Lawanda/PlanetaryDefense/FromitzBoardBase.cs` (the adjacent `NounsForPreciseMatching` fix)
- Proving tests: `UnitTests/MultiNounProcessors/`, `UnitTests/IntentEngine/GiveSomethingToSomeoneDecisionEngineTests.cs`, `Planetfall.Tests/FloydTests.cs`

---

Found via the AdventureBreaker harness (black-box prod + white-box confirmation), discovered during a single Floyd-focused adversarial session. Consolidates AB-048/049/050 (originally filed separately as #362, #352, #353).

arsindelve:
## Summary

Closes #362. Four call sites re-derived a "does the player have/handle this" answer with a narrower, inconsistent check instead of trusting the resolver that had already answered it correctly:

- **DROP resolves to the wrong same-named item** — `TakeOrDropInteractionProcessor.GetItemsToDrop`'s single-item fallbacks resolved through the global, unscoped `Repository.GetItem(noun)`, which could land on a same-named item elsewhere in the game instead of the one actually held (e.g. two Fromitz boards sharing the noun "board"). Now prefers the inventory-scoped `Repository.GetItemInInventory`, falling back to the old global lookup only so `DropIt`'s "you don't have that" messaging still fires correctly for a real-but-unheld item.
- **Floyd goes unresponsive to examine/search/oil while holding an item** — `Floyd.RespondToSimpleInteraction` treated any *non-null* result from a held item's `RespondToSimpleInteraction` as final, but `ItemBase` returns a non-null `NoNounMatchInteractionResult` when the noun doesn't match the held item. Now checks `.InteractionHappened`, matching the pattern already used elsewhere (e.g. `SimpleInteractionEngine`).
- **GIVE/SHOW falsely deny possession of items nested in a worn container** — `GiveSomethingToSomeoneDecisionEngine` and `Floyd.RespondToShow` re-verified possession with a flat `context.Items.Contains(thing)` *after* `Repository.GetItemInScope` had already proved the item reachable by walking the containment hierarchy — so an item nested in a worn/open container (e.g. an ID card in a uniform pocket) was falsely denied. Both now use a new canonical `Repository.IsItemPossessedBy` helper that walks the same hierarchy.
- **`FromitzBoardBase.NounsForPreciseMatching` no-op** — excepted `["card", "access card"]`, copy-pasted from `AccessCard`'s version. Neither string appears in any Fromitz board's noun list, so it was a no-op, leaving "board"/"fromitz board" (the boards' actual generic/ambiguous nouns) eligible for precise matching, contrary to the property's own doc comment.

## Test plan

Each fix has a proving test (TDD: written first, confirmed to fail for the right reason, then fixed to green):

- [x] `Planetfall.Tests/PlanetaryDefenseTests.cs`: `Drop_HeldBoard_DoesNotResolveToDifferentBoardInstance`
- [x] `Planetfall.Tests/PlanetaryDefenseTests.cs`: `GetPreciseMatchInScope_GenericBoardNoun_DoesNotResolve_LeavingDisambiguationToHappen`
- [x] `Planetfall.Tests/FloydTests.cs`: `ExamineFloyd_On_WhileHoldingUnrelatedItem`, `OilFloyd_WhileHoldingUnrelatedItem`, plus a regression guard `ReadDiary_WhileFloydHoldsIt_StillRoutesToDiary`
- [x] `Planetfall.Tests/FloydTests.cs`: `Show_IdCard_ToFloyd_StillInUniformPocket_AsksIfTheyAreUsuallyBlue`
- [x] `UnitTests/IntentEngine/GiveSomethingToSomeoneDecisionEngineTests.cs`: `Should_CallOfferThisThing_When_ItemNestedInsideHeldOpenContainer` (plus fixed 5 pre-existing mock-based tests that didn't set `item.CurrentLocation`, which the new possession check now correctly relies on)
- [x] Full `UnitTests`, `Planetfall.Tests`, `ZorkOne.Tests` suites pass with no regressions
- [x] `dotnet build` on the whole solution succeeds

---
_Generated by [Claude Code](https://claude.ai/code/session_016TToPfcCpcbuzu48XmRzgA)_

### DIFF
--- a/GameEngine/Context.cs
+++ b/GameEngine/Context.cs
@@ -235,7 +235,18 @@ public abstract class Context<T> : IContext where T : IInfocomGame, new()
 
     public void RemoveItem(IItem item)
     {
-        Items.Remove(item);
+        // Try the flat list first, regardless of what item.CurrentLocation currently claims: a
+        // top-level inventory item's CurrentLocation can be stale (e.g. a not-yet-visited room's
+        // lazily-run Init() re-seeding its own starting item singleton overwrites CurrentLocation
+        // without touching this Items list - see LivingRoom's Sword/Lantern). Removing by list
+        // membership first is robust to that and matches this method's original behavior.
+        if (Items.Remove(item))
+            return;
+
+        // Not directly here - it may be nested inside something we hold (e.g. a card in a worn
+        // uniform pocket). Detach it from its actual container so it isn't left duplicated there.
+        if (item.CurrentLocation != this)
+            item.CurrentLocation?.RemoveItem(item);
     }
 
     public void ItemPlacedHere(IItem item)
@@ -400,7 +411,9 @@ public abstract class Context<T> : IContext where T : IInfocomGame, new()
         if (CurrentLocation is not ICanContainItems newLocation)
             throw new Exception("Current location can't hold item");
 
-        Items.Remove(item);
+        // Detach via RemoveItem's flat-list-first logic (robust to a stale CurrentLocation) before
+        // ItemPlacedHere reassigns it - see RemoveItem's comment.
+        RemoveItem(item);
         item.CurrentLocation = newLocation;
         newLocation.ItemPlacedHere(item);
     }
--- a/GameEngine/IntentEngine/GiveSomethingToSomeoneDecisionEngine.cs
+++ b/GameEngine/IntentEngine/GiveSomethingToSomeoneDecisionEngine.cs
@@ -46,7 +46,10 @@ public class GiveSomethingToSomeoneDecisionEngine<TRecipient> where TRecipient :
         if (thing is null)
             return null;
 
-        if (!context.Items.Contains(thing))
+        // Issue #362: GetItemInScope already proved the item is reachable (walking the containment
+        // hierarchy), so re-verify possession the same way rather than with a flat context.Items.Contains
+        // check, which misses items nested inside a worn/open container (e.g. a card in a uniform pocket).
+        if (!Repository.IsItemPossessedBy(thing, context))
             return new PositiveInteractionResult($"You don't have the {thing.NounsForMatching[0]}! ");
 
         return recipient.OfferThisThing(thing, context);
--- a/GameEngine/Item/ItemProcessor/TakeOrDropInteractionProcessor.cs
+++ b/GameEngine/Item/ItemProcessor/TakeOrDropInteractionProcessor.cs
@@ -181,6 +181,11 @@ public class TakeOrDropInteractionProcessor : IVerbProcessor
     {
         if (castItem is null) return new NoNounMatchInteractionResult();
 
+        // Deliberately flat, unlike Repository.IsItemPossessedBy used by GIVE/SHOW: DROP only ever
+        // acts on a top-level held item, even when the resolver above found it nested inside an open
+        // container. Proven by WalkthroughTestTwo's "put leaflet in sack" then "drop leaflet" step,
+        // which expects "You don't have that!" - the original game requires taking a nested item out
+        // before it can be dropped, unlike the (HAVE)-flag reach GIVE/SHOW get into worn containers.
         if (!context.Items.Contains(castItem))
             return new PositiveInteractionResult("You don't have that!");
 
--- a/GameEngine/Item/ItemProcessor/TakeOrDropInteractionProcessor.cs
+++ b/GameEngine/Item/ItemProcessor/TakeOrDropInteractionProcessor.cs
@@ -110,7 +110,12 @@ public class TakeOrDropInteractionProcessor : IVerbProcessor
             // smart enough to match the noun to the item description. An example of this is the "magnet" which is
             // (deliberately, as a puzzle) described as "a metal bar, curved into a U-shape" which the parser does not
             // understand is a magnet. So as a final attempt, let's see if there is a direct noun match.
-            var specificItem = Repository.GetItem(action.Noun);
+            // Issue #362: prefer the scoped inventory match (GetItemInInventory) over the global, unscoped
+            // Repository.GetItem(noun) - the latter can resolve to a same-named item elsewhere in the game
+            // (e.g. a second Fromitz board) instead of the one actually held. Only fall back to the global
+            // lookup when nothing in inventory matches, purely so DropIt can still produce its "you don't
+            // have that" message for a real-but-unheld item instead of a silent no-match.
+            var specificItem = Repository.GetItemInInventory(action.Noun, context) ?? Repository.GetItem(action.Noun);
             return specificItem is not null ? DropIt(context, specificItem) : new NoNounMatchInteractionResult();
         }
 
@@ -118,7 +123,10 @@ public class TakeOrDropInteractionProcessor : IVerbProcessor
         {
             // The AI may return a compound phrase (e.g. "brass lantern") that doesn't resolve;
             // fall back to the noun the IntentParser already extracted from the player's input.
-            var item = Repository.GetItem(items[0]) ?? Repository.GetItem(action.Noun);
+            var item = Repository.GetItemInInventory(items[0], context)
+                       ?? Repository.GetItemInInventory(action.Noun, context)
+                       ?? Repository.GetItem(items[0])
+                       ?? Repository.GetItem(action.Noun);
             return DropIt(context, item);
         }
 
--- a/GameEngine/Item/ItemProcessor/TakeOrDropInteractionProcessor.cs
+++ b/GameEngine/Item/ItemProcessor/TakeOrDropInteractionProcessor.cs
@@ -130,9 +130,12 @@ public class TakeOrDropInteractionProcessor : IVerbProcessor
             return DropIt(context, item);
         }
 
-        // When dropping multiple items, we need to provide feedback for items that don't exist
+        // When dropping multiple items, we need to provide feedback for items that don't exist.
+        // Issue #362: same scoped-first fallback as the single-item branch above - an unscoped
+        // Repository.GetItem(noun) here could resolve "board" to a same-named instance the player
+        // isn't holding, even in a compound "drop board and X" command.
         var itemsWithFeedback = items
-            .Select(itemNoun => (itemNoun, Repository.GetItem(itemNoun)))
+            .Select(itemNoun => (itemNoun, Repository.GetItemInInventory(itemNoun, context) ?? Repository.GetItem(itemNoun)))
             .ToList();
 
         return new PositiveInteractionResult(await DropEverythingProcessor.DropAll(context, itemsWithFeedback, client));
--- a/GameEngine/Repository.cs
+++ b/GameEngine/Repository.cs
@@ -222,13 +222,15 @@ public static class Repository
 
     /// <summary>
     /// Determines whether the player actually possesses the given item - i.e. it is either directly
-    /// in inventory, or nested inside something the player is carrying (a worn/held container's
-    /// contents, or what an <see cref="ICanHoldItems"/> companion is holding for them). This is the
+    /// in inventory, or nested inside a container the player is carrying (worn or held). This is the
     /// canonical possession check: unlike a flat <c>context.Items.Contains(item)</c>, it walks the
     /// containment hierarchy the same way <see cref="IsItemAccessible"/> does, so an item nested
     /// inside a worn container (e.g. an ID card in a uniform pocket) is correctly recognized as held.
     /// Unlike <see cref="IsItemAccessible"/> (which also accepts room-visible items), the chain must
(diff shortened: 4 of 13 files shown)


arsindelve:
Addressed both review comments, plus a duplication bug the fix for #1 exposed:

1. **Multi-item drop scoping** — `GetItemsToDrop`'s `items.Length > 1` branch now resolves through `Repository.GetItemInInventory` first (matching the single-item branch), and `DropEverythingProcessor.DropAll`'s own possession re-check now uses `Repository.IsItemPossessedBy` instead of the flat `context.Items.Contains`.

2. **`IsItemPossessedBy`'s `ICanHoldItems` branch** — confirmed unreachable (Floyd/Troll's `CurrentLocation` is a room, never the player's `IContext`). Removed the dead branch and corrected the doc comment rather than leaving it as a no-op that overstates behavior.

3. **New finding while verifying #2**: loosening the possession check let GIVE/SHOW reach items nested in worn/open containers for the first time — but `Context.RemoveItem`/`Drop` only ever mutated the flat top-level `Items` list, so a nested item (garlic in an open sack, an ID card in a worn uniform pocket) ended up duplicated: referenced by both its real container and wherever it was "moved" to. Fixed `Context.RemoveItem`/`Drop` to detach from the item's actual container, and a related ordering bug in `FloydInventoryManager.OfferItem` that reassigned `CurrentLocation` before calling `RemoveItem`.

Full regression run (`UnitTests`, `Planetfall.Tests`, `ZorkOne.Tests`) is green, including the probabilistic Troll/Thief fight simulators and the spirit-death inventory-scatter test, which briefly regressed to a 0% win rate during iteration on fix #3 before landing on a version that's robust to both the new nested-item case and a separate pre-existing `CurrentLocation` staleness quirk in lazy room `Init()`.

---

arsindelve:
Re: the `DropIt` finding in the review above — investigated this and it turns out the flat check is correct as originally written, not a gap. I reverted my attempt to apply the suggested `Repository.IsItemPossessedBy` fix after it broke the regression suite.

`ZorkOne.Tests/Walkthrough/WalkthroughTestTwo.cs` (a pre-existing test transcribed from a real Zork I transcript) has this sequence: `"put leaflet in sack"` (the sack is open and stays open) followed by `"drop leaflet"`, which asserts the response is `"You don't have that!"`. That's ground truth for the original game's DROP semantics: once an item is nested inside even an *open* container you're holding, DROP requires taking it back out first — it does not reach into the container the way GIVE/SHOW's `(HAVE)` flag does. Applying `IsItemPossessedBy` to `DropIt`/`DropAll` made that walkthrough step (and two cascading steps after it) fail.

So `DropIt` and `DropAll` are intentionally left on the flat `context.Items.Contains` check — added a comment on both citing this evidence, and a new regression test (`Planetfall.Tests/UniformTests.cs`'s `PatrolUniform_DropCard_StillInPocket_DoesNotDrop`) locking in that a nested-but-resolvable item (the ID card in the patrol uniform's open/transparent pocket) correctly reports "You don't have that!" on `drop card` rather than being dropped. Pushed in 4b800bc.

---

PR #368, opened 2026-07-02
Fix #365: stop Floyd from teleporting back into the room mid Bio Lab fight

### LINKED ISSUE #365: Planetfall: Floyd's Bio Lab card-retrieval sequence (BioLockStateMachineManager) never fires in real play — Floyd is never a registered actor in any test
## Bug

Floyd's scripted "I'll get the mini card" Bio Lab sequence — one of Floyd's signature dramatic set pieces, where he offers to rush into the mutated Bio Lab to retrieve the miniaturization access card and (depending on player timing) may sacrifice himself — never fires in real gameplay. `BioLockStateMachineManager.HandleTurnAction` (`Planetfall/Location/Lawanda/Lab/BioLockStateMachineManager.cs`) is completely silent turn after turn even when every documented precondition is met.

## Live reproduction (prod, AdventureBreaker session `floyd-hunt8`)

Single continuous session, no restore involved:

1. In Computer Room with Floyd present and genuinely on, waited until the "computer is broken" concern dialogue fired legitimately (sets `ComputerRoom.FloydHasExpressedConcern = true`).
2. Walked normally (not god-mode) into Bio Lock East: Main Lab → open biolock door → SE → E.
3. Waited 11+ consecutive turns (`wait`, `examine window`, `examine floyd`, a real conversation with Floyd) — moves 282 through 292.
4. Expected: after the first "silent" turn, the very next turn should say (`FloydConstants.NeedToGetCard`): *"Floyd stands on his tiptoes and peers in the window. 'Looks dangerous in there,' says Floyd... We'll need card there to fix computer... Floyd will get card... You open the door, then Floyd will rush in..."*
5. Actual: only `"Time passes..."` on every turn. The dialogue never appeared once.

Sanity checks performed to rule out the obvious causes:
- `floyd, look at the card` → Floyd gave a real AI-generated response, proving `Floyd.IsOn == true` (per `OnBeingTalkedTo`'s explicit off-check).
- `god mode where floyd` → reported `Bio Lock East`, confirming `floyd.CurrentLocation` correctly matches `context.CurrentLocation` (rules out a stale-reference/location-mismatch explanation).
- `HandleTurnAction`'s only gate before the dialogue branches is `if (!isFloydInLab && !isFloydHereAndOn) return string.Empty;` — both `isFloydInLab` (irrelevant here, sequence hadn't started) and `isFloydHereAndOn` were independently confirmed true, yet the dialogue still never fired across 11+ turns.
- The engine's top-level exception safety net (issue #271, `GameEngine.HandleUnexpectedEngineError`) replaces the *entire* turn response with an AI "oops" narration on any unhandled exception — but every turn returned the exact expected `"Time passes..."` text with nothing appended, ruling out a silently-caught exception.

## Root cause (test-suite audit)

`BioLockEast` and `Floyd` are two *independent* `ITurnBasedActor`s. In real gameplay both end up registered in `Context.Actors` and both run every turn inside `GameEngine.ProcessActors()`'s `foreach` loop:

- `BioLockEast.BeforeEnterLocation` calls `context.RegisterActor(this)` on entry (normal).
- `Floyd` is registered via `FloydPowerManager.Activate()` — **the only** code path that calls `context.RegisterActor(floyd)` — which fires whenever a player actually types `activate floyd` / `turn on floyd`. Every real playthrough reaches Bio Lock East with Floyd already activated this way.

But **every test that exercises this state machine sets `floyd.IsOn = true` by direct property assignment**, never by calling real activation:

- `Planetfall.Tests/BioLockEastTests.cs` — every single test (`Floyd_ComputerRoomConcernExpressed_ShouldSayNeedToGetCard`, etc.) does `floyd.IsOn = true;` then only `target.Context.RegisterActor(bioLockEast);` — Floyd is *never* added to `Context.Actors`.
- `Planetfall.Tests/Walkthrough/WalkthroughBioLock.cs`'s `SetupBioLock()` helper does the same: `bioLockEast.ItemPlacedHere(floyd); floyd.IsOn = true;` with no call to Floyd's real activation path either.

So **no test in the suite — including the end-to-end walkthrough — ever has Floyd actually registered as an actor at the same time as BioLockEast.** This is presumably not an oversight: `BioLockEastTests.cs` has an explicit comment admitting the gap was found and routed around rather than fixed:

```csharp
// Complete sequence test removed - individual behaviors are tested above
// The integration has an issue where both Floyd and BioLockEast act on the same turn
// causing double-incrementing of counters. Individual tests verify correctness.
```

That "double-incrementing" issue is presumably the same interaction that's now silently killing the dialogue entirely in production. I was not able to run the C# test suite locally in this session (no dotnet SDK in the AdventureBreaker harness environment used for this playtest) to pin the exact line-level mechanism inside `ProcessActors()`'s actor-ordering/interaction — but the live prod reproduction is solid and repeatable, and the test-suite gap independently corroborates that this exact production configuration (both actors registered, Floyd for real) has never actually been exercised.

## Impact

A player following the intended path (get Floyd concerned about the computer, walk to the Bio Lab window) gets **no in-fiction prompt at all** that they should open the door. The only way to discover the sequence exists is to guess and open the door blind — which instead hits `BioLockStateMachineManager.HandleDoorOpening`'s fallback branch and kills the player instantly:

> "Opening the door reveals a Bio-Lab full of horrible mutations. You stare at them, frozen with horror. Growling with hunger and delight, the mutations march into the bio-lock and devour you."

This makes one of Floyd's signature dramatic beats — and the miniaturization-access-card puzzle it gates — effectively unreachable via normal play.

## Suggested next step

Add a test (or fix the walkthrough setup) that registers Floyd as a real actor alongside `BioLockEast` — e.g. via `FloydPowerManager.Activate(context)` instead of `floyd.IsOn = true` — and trace what actually happens in `ProcessActors()`'s loop when both act the same turn. That should surface the exact mechanism the removed "complete sequence" integration test apparently hit.

---
Found via AdventureBreaker adversarial playtesting (`floyd-hunt8` session). Tracked as `AB-052` in the AdventureBreaker coverage ledger.

arsindelve:
## Summary

Fixes #365. The bug report's own hypothesis - "no test ever registers Floyd as a real actor alongside `BioLockEast`" - turned out to be the exact right lead, but the actual defect it exposes is more specific than "the dialogue never fires."

- In real play, Floyd is registered as an `ITurnBasedActor` via `FloydPowerManager.Activate()` long before the player ever reaches Bio Lock East. Because he's added to `Context.Actors` first, his own `Act()` always runs *before* `BioLockEast.Act()` in `ProcessActors()`'s loop each turn.
- `BioLockStateMachineManager.HandleTurnAction` calls `floyd.SkipActingThisTurn(context)` from several branches, intending to keep Floyd quiet while he's "in the lab" - but since `BioLockEast.Act()` always runs *after* Floyd's own `Act()` in the same turn, that call always arrives one actor-turn too late to do anything (and `ProcessBeginningOfTurn` resets the flag again before Floyd's next turn regardless).
- When `HandleDoorOpening` sets `floyd.CurrentLocation = null` to represent Floyd being absent (fighting mutants behind the closed door), Floyd's own `FloydMovementManager.HandleFollowingPlayer` sees the location mismatch on his very next `Act()` and "helpfully" teleports him straight back into the room, printing `"Floyd follows you."` - while the narration in the same breath says he's trapped fighting for his life on the other side of the door. This also puts him back in `BioLockEast.Items`, contradicting the sequence's own invariant (checked elsewhere by `WhileFloydFighting_FloydNotInRoomItems`).

Every existing `BioLockEastTests`/`WalkthroughBioLock` test only ever calls `Context.RegisterActor(bioLockEast)`, never `Context.RegisterActor(floyd)`, so Floyd's `Act()` never actually ran and this gap was invisible to the suite - exactly as the issue predicted.

### Fix

- Added `Floyd.IsAwayOnScriptedSequence`, mirroring the existing `IsOffWandering` guard.
- `BioLockStateMachineManager` sets it when Floyd rushes into the lab (alongside the existing `CurrentLocation = null`).
- `FloydMovementManager.HandleFollowingPlayer` now treats it the same as `IsOffWandering`: don't try to reposition Floyd back into the room.

### Follow-up fixes (from a recall-focused `/code-review` of the original fix)

The initial fix introduced two regressions of its own, both now fixed with regression tests:

1. **Floyd could be permanently softlocked.** `IsAwayOnScriptedSequence` was set `true` but never reset. If the player abandons the sequence mid-fight by walking away instead of closing the door, `BioLockEast` stops acting (the state machine freezes) but Floyd stays parked with no location - forever, since nothing else ever clears the flag. `examine floyd` returned a blank response, with no in-game recovery. Fixed by releasing the flag in `BioLockEast.OnLeaveLocation` when the sequence is abandoned mid-flight - this also fixes a "ghost talk" case where Floyd could narrate being active in a room he couldn't physically be in.
2. **Door-reopen didn't actually place Floyd back in the room.** `HandleDoorOpening`'s `NeedToReopenDoor` branch narrates "Floyd stumbles out of the Bio Lab, clutching the mini-booth card" but never called `ItemPlacedHere` - a pre-existing bug, but the original fix made it 100% reproducible instead of ~80% accidentally masked by the old buggy auto-follow it removed.

The review also surfaced two more severe but genuinely pre-existing and unrelated bugs in the same state machine (confirmed present with or without this PR's changes): opening the door and then simply walking out and back in leaves a turn counter un-reset, causing an instant unavoidable player death on re-entry; and a god-mode teleport away during the same window causes an unrelated death later regardless of player location. Left out of this PR's scope since they're a different root cause (state never resets when the sequence is abandoned via normal movement, independent of the actor-registration bug this PR fixes) - worth a follow-up issue.

## Test plan

- [x] `Planetfall.Tests/BioLockEastTests.cs`: `CloseDoor_WithFloydRegisteredAsRealActor_FloydShouldRemainAbsentWhileFighting` - registers both Floyd and `BioLockEast` as real actors (matching production), verified failing before the fix and passing after.
- [x] `AbandoningSequence_ByWalkingAway_ShouldNotPermanentlySoftlockFloyd` - proves the softlock regression and its fix.
- [x] `ReopenDoor_AfterInTheLabFour_FloydShouldActuallyBeBackInTheRoom` - proves the missing-placement bug and its fix.
- [x] `dotnet test Planetfall.Tests` - 2859 passed, 0 failed.
- [x] `dotnet test UnitTests` - 985 passed, 0 failed.
- [x] `dotnet test ZorkOne.Tests` - 1750 passed, 0 failed.
- [x] Manually re-verified the full `WalkthroughBioLock` 48-scenario matrix with Floyd genuinely registered as an actor (temporary local edit, not committed) to confirm the whole sequence - waiting, opening/closing the door, Floyd rushing in, fighting, dying, and the successful sacrifice ending - still narrates correctly end-to-end.

### DIFF
--- a/Planetfall.Tests/BioLockEastTests.cs
+++ b/Planetfall.Tests/BioLockEastTests.cs
@@ -398,6 +398,41 @@ public class BioLockEastTests : EngineTestsBase
     }
 
     [Test]
+    public async Task AbandoningSequence_ByWalkingAway_ShouldNotPermanentlySoftlockFloyd()
+    {
+        // Issue #365 follow-up: IsAwayOnScriptedSequence is only ever set true (never cleared) by the
+        // Bio Lab sequence. If the player walks away instead of closing the door, BioLockEast unregisters
+        // itself as an actor (OnLeaveLocation) and the state machine freezes forever - so without an
+        // explicit recovery, Floyd would stay CurrentLocation=null and IsAwayOnScriptedSequence=true for
+        // the rest of the game, with no in-game way to get him back. BioLockEast.OnLeaveLocation must
+        // release the flag when the sequence is abandoned mid-flight, letting Floyd's normal following
+        // behavior resume.
+        var target = GetTarget();
+        var floyd = GetItem<Floyd>();
+        floyd.IsOn = true;
+        floyd.TurnOnCountdown = 0;
+        floyd.HasEverBeenOn = true;
+        var mockChooser = new Mock<IRandomChooser>();
+        mockChooser.Setup(r => r.RollDiceSuccess(5)).Returns(false); // deterministic: always the "follow" branch
+        floyd.Chooser = mockChooser.Object;
+        var bioLockEast = StartHere<BioLockEast>();
+        bioLockEast.ItemPlacedHere(floyd);
+        bioLockEast.StateMachine.HasWaitedOneTurnInBioLockEast = true;
+        bioLockEast.StateMachine.FloydHasSaidNeedToGetCard = true;
+        target.Context.RegisterActor(floyd);
+        target.Context.RegisterActor(bioLockEast);
+
+        await target.GetResponse("open door"); // Floyd rushes in: CurrentLocation=null, IsAwayOnScriptedSequence=true
+        floyd.IsAwayOnScriptedSequence.Should().BeTrue();
+
+        var response = await target.GetResponse("west"); // Abandon the sequence instead of closing the door
+
+        floyd.IsAwayOnScriptedSequence.Should().BeFalse("leaving mid-sequence must release Floyd, not strand him forever");
+        response.Should().Contain("Floyd follows you");
+        floyd.CurrentLocation.Should().NotBeNull("Floyd must be recoverable, not permanently stuck with no location");
+    }
+
+    [Test]
     public async Task FloydFighting_Turn1_ShouldShowInTheLabTwo()
     {
         var target = GetTarget();
@@ -467,6 +502,28 @@ public class BioLockEastTests : EngineTestsBase
         await target.GetResponse("open door");
 
         bioLockEast.StateMachine.LabSequenceState.Should().Be(FloydLabSequenceState.DoorReopenedNeedToCloseAgain);
+    }
+
+    [Test]
+    public async Task ReopenDoor_AfterInTheLabFour_FloydShouldActuallyBeBackInTheRoom()
+    {
+        // The narration ("Floyd stumbles out of the Bio Lab, clutching the mini-booth card") says he's
+        // back, but earlier in the sequence he was correctly given CurrentLocation = null while fighting
+        // (unlike sibling tests here, which pre-seed ItemPlacedHere and so never exercise this transition
+        // starting from the real null-location state). Reopening the door must actually restore his
+        // presence, not just narrate it.
+        var target = GetTarget();
+        var floyd = GetItem<Floyd>();
+        floyd.IsOn = true;
+        floyd.CurrentLocation = null;
+        var bioLockEast = StartHere<BioLockEast>();
+        bioLockEast.StateMachine.LabSequenceState = FloydLabSequenceState.NeedToReopenDoor;
+        target.Context.RegisterActor(bioLockEast);
+
+        await target.GetResponse("open door");
+
+        floyd.CurrentLocation.Should().Be(bioLockEast);
+        bioLockEast.Items.Should().Contain(floyd);
     }
 
     [Test]
--- a/Planetfall.Tests/BioLockEastTests.cs
+++ b/Planetfall.Tests/BioLockEastTests.cs
@@ -1,4 +1,6 @@
 ﻿using FluentAssertions;
+using Model.Interface;
+using Moq;
 using Planetfall.Item.Kalamontee.Mech.FloydPart;
 using Planetfall.Item.Lawanda.Lab;
 using Planetfall.Location.Lawanda;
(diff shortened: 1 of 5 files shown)


Planetfall.Tests/BioLockEastTests.cs
```diff
@@ -358,6 +360,43 @@ public async Task CloseDoor_AfterFloydRushesIn_ShouldShowDoorClosedMessage()
         bioLockEast.StateMachine.LabSequenceState.Should().Be(FloydLabSequenceState.DoorClosedFloydFighting);
     }
 
+    [Test]
+    public async Task CloseDoor_WithFloydRegisteredAsRealActor_FloydShouldRemainAbsentWhileFighting()
```
arsindelve, line 364: Is this sufficient test coverage?

arsindelve: Good question — it wasn't. I ran a recall-focused review of this fix and it turned up two real regressions from registering Floyd as a genuine actor, both now fixed in 192c218 with new tests:

1. **Floyd could be permanently softlocked.** `IsAwayOnScriptedSequence` was set `true` when Floyd rushes into the lab but never reset. If the player abandons the sequence by walking away instead of closing the door, `BioLockEast` stops acting (state machine frozen) but Floyd stays parked with no location forever — `examine floyd` returned blank, with no in-game recovery. Fixed by releasing the flag in `BioLockEast.OnLeaveLocation` when the sequence is abandoned mid-flight (also incidentally fixes a "ghost talk" case where Floyd could narrate being active in a room he couldn't physically be in). New test: `AbandoningSequence_ByWalkingAway_ShouldNotPermanentlySoftlockFloyd`.

2. **Door-reopen didn't actually place Floyd back in the room.** `HandleDoorOpening`'s `NeedToReopenDoor` branch narrates "Floyd stumbles out of the Bio Lab, clutching the mini-booth card" but never called `ItemPlacedHere` — pre-existing bug, but this PR made it 100% reproducible instead of ~80% accidentally masked by the old buggy auto-follow. New test: `ReopenDoor_AfterInTheLabFour_FloydShouldActuallyBeBackInTheRoom`.

The review also surfaced two more severe, but genuinely pre-existing and unrelated bugs in the same state machine (confirmed unaffected by this diff either way): opening the door and then simply walking out and back in leaves `TurnsAfterFloydRushedIn` un-reset, causing an instant unavoidable player death on re-entry; and a god-mode teleport away during the same window causes an unrelated death later regardless of player location. Left those out of this PR's scope since they're a different root cause — happy to file a follow-up issue for them if you'd like.

---

arsindelve:
Thanks for the review. Addressed the actionable points in 216cf26: added a one-line comment on the `NeedToReopenDoor` branch explaining why `IsAwayOnScriptedSequence` is deliberately left set there, and updated the PR description to call out the reopen-placement fix as distinct from the main teleport-back bug.

One clarification on the "PerformRandomAction not gated" point: that path requires `context.CurrentLocation` to be outside `BioLockEast` (inside it, `IFloydDoesNotTalkHere` blocks `PerformRandomAction` unconditionally regardless of any flag), which only happens once the player has left the room - and `BioLockEast.OnLeaveLocation` clears `IsAwayOnScriptedSequence` synchronously as part of that same move, before `Floyd.Act()` runs later that turn. I verified this empirically (forced `PerformRandomAction`'s dice to hit and drove "open door" → "west"): Floyd's location is restored before his `Act()` ever reaches that branch, so no ghost chatter. So I believe this one's already covered by the `OnLeaveLocation` fix rather than a separate gap - flag me if you were picturing a different path to it.

The `TurnsAfterFloydRushedIn`-not-reset-on-re-entry point is real and matches what I found independently (confirmed present with or without this PR's changes) - agreed it's a separate, pre-existing issue outside #365's scope. I'll leave it for a follow-up issue rather than scope-creep this PR further.

---

Issue #374, opened 2026-07-02
Add a "god mode kill &lt;creature&gt;" command for deterministic playtesting past combat gates (starting with the Zork I troll)


arsindelve:
## Motivation

The Troll Room is a mandatory gate to a large chunk of Zork I content (the entire Maze, and everything reachable through it — the Cyclops Room, the Thief's Treasure Room, the Grating Room exit) — the troll blocks both the east and west passages until defeated. Troll combat is randomized and genuinely punishing: reading `AdventurerVersusTrollCombatEngine.cs` and `TrollCombatEngine.cs`, the troll registers as a turn-based actor *during room entry* (`TrollRoom.BeforeEnterLocation`, line 75: `context.RegisterActor(GetItem<Troll>())`), so it gets an unprovoked attack the same turn the player walks in, before the player's first real action — and that attack alone has roughly a 16% instant-kill chance per the `_outcomes`/`_haveWeaponOutcomes` combat tables (3 Fatal outcomes out of 19 total weighted entries when armed).

While adversarially playtesting the maze via the AdventureBreaker harness (black-box prod), this cost real time: **5 deaths out of 7 attempts** just to get past the troll and reach the area under test, with several deaths happening before the player got to act at all. The combat probabilities themselves look reasonably calibrated (not an obvious bug — see the discussion in this session), but there's currently no way to skip this gate deterministically for testing purposes downstream of it, the way `god mode reset time` and `god mode no survival` already let testers skip Planetfall's chronometer/survival-clock gates.

## Request

A `god mode kill <creature>` command, following the existing `GodModeProcessor.cs` conventions (`god mode take <item>`, `god mode go <place>`, `god mode where <item>`, `god mode no survival`, `god mode reset time`):

- Add a new dispatch branch: `if (input.Contains(" kill ")) return Task.FromResult(Kill(input, context));`
- Resolve the target via the same `Repository.GetItem(lastWord)` / `GetWordsAfterTarget` pattern `Take`/`Go`/`Where` already use.
[...]

PR #375, opened 2026-07-02
Add "god mode kill &lt;creature&gt;" for deterministic playtesting past the troll (#374)

arsindelve:
## Summary

- Adds a `god mode kill <creature>` command to `GodModeProcessor`, following the existing `god mode take/go/where` conventions, so playtesting can skip the Zork I Troll Room's randomized (and punishing) combat gate deterministically.
- `ICanBeAttacked` (`Model/Item/ICanBeAttacked.cs`) gains a `GodModeKill()` default-interface method that returns `false` by default. This lets `GodModeProcessor` (in `GameEngine`, which doesn't reference `ZorkOne`) dispatch through the shared `Model`-level interface without a layering violation, while existing implementers (`Thief`, `Cyclops`) keep compiling unchanged.
- `Troll` overrides `GodModeKill()` to mirror `AdventurerVersusTrollCombatEngine.DeathBlow`: sets `IsDead = true`, drops the axe into the room via `ItemPlacedHere`, and clears `CurrentLocation` — the same end state a real fatal blow produces.
- Per the issue, this first pass is scoped to the Troll (the actual pain point); `Thief`/`Cyclops` keep the no-op default and can opt in later by overriding `GodModeKill()`.
- Unresolvable/non-`ICanBeAttacked` targets fall through to the existing generic `"Invalid use of God mode. Bad adventurer!"` message, matching `Take`'s behavior.

## Test plan

- [x] Added a failing test first (`GodModeKill_Troll_KillsHimAndDropsTheAxe`), confirmed it fails for the right reason (`troll.IsDead` was `False`) with the production code reverted, then restored the fix and confirmed it passes (TDD red/green).
- [x] Added `GodModeKill_UnrecognizedTarget_IsInvalid` covering the fallback path.
- [x] `dotnet test UnitTests/UnitTests.csproj` — 995 passed, 1 skipped (pre-existing, unrelated).
- [x] `dotnet test ZorkOne.Tests/ZorkOne.Tests.csproj` — 1752 passed.
- [x] `dotnet test Planetfall.Tests/Planetfall.Tests.csproj` — 2904 passed (verifies the shared `ICanBeAttacked` interface change doesn't regress Planetfall).

https://claude.ai/code/session_01KtHXYEbwH6zdAFhc2h9PKP

---
_Generated by [Claude Code](https://claude.ai/code/session_01KtHXYEbwH6zdAFhc2h9PKP)_


ZorkOne/Item/Cyclops.cs
```diff
@@ -146,4 +146,27 @@ public override string GenericDescription(ILocation? currentLocation)
...
         return NeverPickedUpDescription(currentLocation);
     }
+
+    /// <summary>
+    /// Issue #374: shared removal logic used by both "say Ulysses" (CyclopsRoom's flee handler) and
+    /// "god mode kill cyclops". The cyclops has no combat-death path in this engine - the only way
+    /// he ever leaves is fleeing the room - so "kill" reuses that same removal rather than inventing
+    /// a separate, divergent mechanism.
+    /// </summary>
+    internal void Die(IContext context)
```
arsindelve: Extract this out to a base class for all three

arsindelve: Done in 422da67. Added `CombatGateCreatureBase` (`: ContainerBase, ICanBeAttacked`) that owns `IsDead`, the once-only guard, and the `GodModeKill` wrapper; Troll/Thief/Cyclops now only implement `OnDeath(context)` for their own removal effects.

One note: this required moving `Cyclops` off `ItemBase` onto `CombatGateCreatureBase` (via `ContainerBase`), since Troll/Thief already needed `ContainerBase`'s `Items` list and there's no way to share one base across an `ItemBase`-rooted type and a `ContainerBase`-rooted type otherwise. I checked the two behavioral side effects that move could cause:
- `ContainerBase.HasMatchingNoun` is stricter (exact match) than `ItemBase`'s word-boundary fallback — but real gameplay parsing strips articles upstream before it ever reaches that check, so `"kill the cyclops with the sword"` still passes (existing `CyclopsTests` cover this). Only the raw `god mode kill <noun>` path is affected, and that's the same pre-existing "no article" limitation Troll already had.
- Cyclops now technically accepts `ICanContainItems` puts (`put X in cyclops`) — untested before and after, but Troll and Thief already had this from `ContainerBase` with no guard, so this just makes Cyclops consistent with its siblings rather than introducing something new.

Full suite (UnitTests/ZorkOne.Tests/Planetfall.Tests) green.

---

PR #380, opened 2026-07-03
Fix #369: jumping into the rift now triggers the written death scene

### LINKED ISSUE #369: Planetfall: jumping into the rift never triggers the written death — always misrouted to the generic "too wide to jump across" movement failure
## Bug

`RiftLocationBase` (shared by Admin Corridor and Admin Corridor North, either side of the rift) has a deliberately-written death scene for jumping into the rift, but it is **unreachable in live play** — every natural phrasing a player would use instead falls through to the mundane "too wide to jump across" movement-failure message.

Confirmed live on prod, both `play` (narrator on) and `quiet` (narrator off, deterministic) mode, from both `AdminCorridor` and `AdminCorridorNorth`:

- `jump into rift` → "The rift is too wide to jump across."
- `jump over rift` → "The rift is too wide to jump across."
- `jump rift` (bare noun, no preposition at all) → "The rift is too wide to jump across."
- `enter rift` → "The rift is too wide to jump across."

None of these ever produce the death text that's sitting right there in the code. By contrast, a bare `jump` (no noun) correctly does *not* trigger death — it gets a generic "you hop in place" narration — confirming the routing issue is specific to verb+rift-noun combinations, not jump in general.

## Root cause

`Planetfall/Location/Kalamontee/Admin/RiftLocationBase.cs:31-34`:

```csharp
public override async Task<InteractionResult> RespondToSimpleInteraction(SimpleIntent action, IContext context,
    IGenerationClient client, IItemProcessorFactory itemProcessorFactory)
{
    if (action.Match(Verbs.JumpVerbs, RiftNouns))
        return new DeathProcessor().Process(
            "You get a brief (but much closer) view of the sharp and nasty rocks at the bottom of the rift. ",
            context);
    ...
```

This branch only fires for a `SimpleIntent{Verb ∈ JumpVerbs, Noun ∈ RiftNouns}`. But `AdminCorridor.Map()` / `AdminCorridorNorth.Map()` both wire `Direction.N` / `Direction.S` toward the rift with:

```csharp
// AdminCorridor.cs:16-22 (mirrored in AdminCorridorNorth.cs)
Direction.N,
new MovementParameters
{
    CanGo = _ => GetItem<Ladder>().IsAcrossRift,
    CustomFailureMessage = "The rift is too wide to jump across. ",
    Location = Repository.GetLocation<AdminCorridorNorth>()
}
```

"The rift is too wide to jump across." only exists in this one `CustomFailureMessage` string — so seeing it for `jump rift`/`jump into rift`/etc. proves those inputs are being classified as a `MoveIntent{Direction=N}` and routed through `MoveEngine.Process` (`GameEngine.cs:737`), never reaching `RiftLocationBase.RespondToSimpleInteraction` at all. The live AI-based complex-intent parser (`IntentParser.DetermineComplexIntentType` → `AskTheAIParser`) is semantically (and reasonably, from a pure-NLU standpoint) interpreting "jump toward the thing to the north" as an attempt to move north, rather than as a `SimpleIntent` verb+noun action — but that classification silently swallows the location's own, more specific, already-written handler.

I don't have ZIL ground truth available in this environment to confirm the original's exact JUMP handling, but that's not really the crux here: this isn't a "faithful cruelty vs. bug" judgment call — the C# already contains a fully-written, deliberate death scene for this exact action, and it is provably unreachable through any phrasing tested. That's a routing/parsing defect regardless of what the original did.

## Test coverage gap

Zero tests anywhere in `Planetfall.Tests` exercise the jump+rift death branch. The one existing test that asserts "too wide to jump" (`AdminCorridorTests.CannotGoNorth_WithoutLadder`) uses bare direction input `"n"`, not `"jump"` — so the suite doesn't accidentally codify this bug as intended behavior; the branch is just genuinely never exercised.

## Impact

Medium. Jumping into a lethal chasm is a classic "let's see what happens" player action in text adventures, and Planetfall is well known for punishing exactly this kind of curiosity. Right now that curiosity is met with an anticlimactic, contextually-wrong movement-failure message instead of the vivid death scene the game clearly intends. Not blocking — the rift is still crossable via the ladder — but it's dead content and a misleading response for a natural, expected input.

## Suggested fix

The production fix lives in intent classification, not in `RiftLocationBase` itself (that code is already correct). Options:
- A deterministic pre-check (mirroring `DirectionParser.IsDirection`'s early-out for `"enter "`/`"exit "` prefixes) that recognizes `<JumpVerb> <noun-matching-current-room-scenery>` and forces a `SimpleIntent` before movement inference gets a chance to claim it.
- Or tune the AI parser's prompt/priority so an explicit scenery noun (not just a bare direction-shaped phrase) wins over movement inference.

## TDD plan

`TestParser.cs` (the deterministic unit-test parser) doesn't handle `"jump"` as a verb at all today (`"jump"` isn't in its `_verbs` list), so `"jump rift"` currently resolves to `NullIntent` in tests rather than reproducing the live `MoveIntent` misroute. Suggested red-before/green-after:
1. In `TestParser.cs`, add a deterministic case for `"jump"/"leap" + <rift noun>"` (mirroring the existing `"go through "` special-case at lines 158-169) so it reliably yields `SimpleIntent{Verb="jump", Noun="rift"}`.
2. Add a failing-first test to `Planetfall.Tests/AdminCorridorTests.cs` (alongside `CannotGoNorth_WithoutLadder`) asserting `GetResponse("jump rift")` (and `"jump into rift"`) contains the death text / "You have died", using the same `StartHere<AdminCorridor>()` setup pattern already in that file.
3. This proves the location handler itself works in isolation; the live-parser fix (see above) is the part that actually makes the death reachable in production and isn't unit-testable without mocking the real AI response.

---
Found via AdventureBreaker adversarial playtesting (black-box prod, confirmed identically in both `play` and `quiet` mode; white-box root-caused against `../ZorkAI` — no ZIL checkout available this session to cross-check original intent).

arsindelve:
## Summary

Fixes #369. Jumping into the Admin Corridor rift (Planetfall) has a deliberately-written death scene in `RiftLocationBase.RespondToSimpleInteraction`, but it was unreachable in live play: "jump rift", "jump into rift", and "jump over rift" were all misclassified by the complex-intent (AI) parser as movement toward the rift's `Direction.N` exit, so they only ever produced the mundane "The rift is too wide to jump across." movement-failure message from `AdminCorridor`/`AdminCorridorNorth`'s `Map()` instead of the death text.

- `RiftLocationBase` now overrides `RespondToSpecificLocationInteraction` to recognize jump-verb + rift-noun phrasing (with optional filler prepositions/articles: "into", "in", "over", "across", "through", "the") directly from the raw player input. This runs *before* any parser — deterministic or AI — gets a chance to claim the input, so the existing death branch actually fires, in both tests and production.
- The original `RespondToSimpleInteraction` death branch is left untouched as a fallback for any `SimpleIntent` that reaches it through another path.

## Test plan (TDD)

- [x] Added failing-first tests in `Planetfall.Tests/AdminCorridorTests.cs` (`JumpIntoRift` nested fixture) covering: bare "jump rift", "jump into rift", "jump over rift", a rift-noun synonym ("jump into the chasm"), the same fix from `AdminCorridorNorth`, and a regression guard that bare "jump" (no noun) still does not kill the player.
- [x] Confirmed all 5 new death-scene tests failed for the right reason (empty response, no death text) before the fix.
- [x] Confirmed all tests pass after the fix.
- [x] Ran `dotnet test Planetfall.Tests` (2922 passed, 0 failed) — no regressions.
- [x] Ran `dotnet test UnitTests` (993 passed, 0 failed) — no regressions.
- [x] Ran `dotnet test ZorkOne.Tests` (1752 passed, 0 failed) — no regressions.
- [x] Full solution build succeeds with 0 errors.

https://claude.ai/code/session_01CREk8m265rZVK6dNt9vrYP

---
_Generated by [Claude Code](https://claude.ai/code/session_01CREk8m265rZVK6dNt9vrYP)_

### DIFF
--- a/Planetfall/Location/Kalamontee/Admin/RiftLocationBase.cs
+++ b/Planetfall/Location/Kalamontee/Admin/RiftLocationBase.cs
@@ -39,4 +39,37 @@ internal abstract class RiftLocationBase : LocationWithNoStartingItems
 
         return await base.RespondToSimpleInteraction(action, context, client, itemProcessorFactory);
     }
+
+    // Issue #369: "jump rift" / "jump into rift" / "jump over rift" never reached the death branch
+    // above. The AI complex-intent parser reads the rift as "the thing to the north" and misclassifies
+    // these as a MoveIntent toward Direction.N, which only ever produces the mundane "too wide to jump
+    // across" movement-failure message from AdminCorridor/AdminCorridorNorth's Map(). Catching the raw
+    // input here - RespondToSpecificLocationInteraction runs before any parser (deterministic or AI)
+    // gets a chance to claim it - lets the already-correct death scene above actually fire.
+    public override async Task<InteractionResult> RespondToSpecificLocationInteraction(string? input,
+        IContext context, IGenerationClient client)
+    {
+        if (IsJumpingIntoTheRift(input))
+            return new DeathProcessor().Process(
+                "You get a brief (but much closer) view of the sharp and nasty rocks at the bottom of the rift. ",
+                context);
+
+        return await base.RespondToSpecificLocationInteraction(input, context, client);
+    }
+
+    private static bool IsJumpingIntoTheRift(string? input)
+    {
+        var words = input?.ToLowerInvariant().Trim().Split(' ', StringSplitOptions.RemoveEmptyEntries).ToList();
+        if (words is null || words.Count == 0 || !Verbs.JumpVerbs.Contains(words[0]))
+            return false;
+
+        words.RemoveAt(0);
+
+        // Drop the filler prepositions/articles a player might naturally insert - "jump into the
+        // rift" and "jump over rift" mean the same thing as the bare "jump rift".
+        string[] filler = ["into", "in", "over", "across", "through", "the"];
+        words.RemoveAll(filler.Contains);
+
+        return RiftNouns.Contains(string.Join(' ', words));
+    }
 }
--- a/Planetfall.Tests/AdminCorridorTests.cs
+++ b/Planetfall.Tests/AdminCorridorTests.cs
@@ -97,6 +97,81 @@ public class AdminCorridorTests : EngineTestsBase
         Context.CurrentLocation.Should().BeOfType<AdminCorridor>();
     }
 
+    // Issue #369: "jump rift"/"jump into rift"/"jump over rift" were being misrouted to the
+    // Direction.N movement failure ("too wide to jump across") instead of reaching the deliberate
+    // death scene already written in RiftLocationBase.RespondToSimpleInteraction. A bare "jump" with
+    // no rift noun must still fall through to normal (non-fatal) handling.
+    [TestFixture]
+    public class JumpIntoRift : AdminCorridorTests
+    {
+        [Test]
+        public async Task JumpRift_TheBareNoun_DiesInTheRift()
+        {
+            var target = GetTarget();
+            StartHere<AdminCorridor>();
+
+            var response = await target.GetResponse("jump rift");
+
+            response.Should().Contain("You have died");
+            response.Should().Contain("sharp and nasty rocks");
+        }
+
+        [Test]
+        public async Task JumpIntoRift_DiesInTheRift()
+        {
+            var target = GetTarget();
+            StartHere<AdminCorridor>();
+
+            var response = await target.GetResponse("jump into rift");
+
+            response.Should().Contain("You have died");
+        }
+
+        [Test]
+        public async Task JumpOverRift_DiesInTheRift()
+        {
+            var target = GetTarget();
+            StartHere<AdminCorridor>();
+
+            var response = await target.GetResponse("jump over rift");
+
+            response.Should().Contain("You have died");
+        }
+
+        [Test]
+        public async Task JumpIntoTheChasm_SynonymNounDiesInTheRift()
+        {
+            var target = GetTarget();
+            StartHere<AdminCorridor>();
+
+            var response = await target.GetResponse("jump into the chasm");
+
+            response.Should().Contain("You have died");
+        }
+
+        [Test]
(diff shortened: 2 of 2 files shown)


Planetfall/Location/Kalamontee/Admin/RiftLocationBase.cs
```diff
@@ -39,4 +39,37 @@ public override async Task<InteractionResult> RespondToSimpleInteraction(SimpleI
...
         return await base.RespondToSimpleInteraction(action, context, client, itemProcessorFactory);
     }
+
+    // Issue #369: "jump rift" / "jump into rift" / "jump over rift" never reached the death branch
+    // above. The AI complex-intent parser reads the rift as "the thing to the north" and misclassifies
+    // these as a MoveIntent toward Direction.N, which only ever produces the mundane "too wide to jump
+    // across" movement-failure message from AdminCorridor/AdminCorridorNorth's Map(). Catching the raw
+    // input here - RespondToSpecificLocationInteraction runs before any parser (deterministic or AI)
+    // gets a chance to claim it - lets the already-correct death scene above actually fire.
+    public override async Task<InteractionResult> RespondToSpecificLocationInteraction(string? input,
```
arsindelve, line 49: What about from the north?

arsindelve: Already covered — `RespondToSpecificLocationInteraction` is defined on `RiftLocationBase` itself, which both `AdminCorridor` and `AdminCorridorNorth` inherit from, so the same jump-verb + rift-noun check applies on both sides of the rift. There's a dedicated test for the north side: `AdminCorridorTests.JumpIntoRift.JumpFromAdminCorridorNorth_DiesInTheRift` (starts at `AdminCorridorNorth`, sends "jump into rift", asserts death). I re-ran it locally just now and it passes.

---

Issue #423, opened 2026-07-21
Bio Lock East: "look through window" shows the room instead of the Bio Lab view (mutants + mini card), though the handler explicitly tries to support it


arsindelve:
## Bug

At **Bio Lock East**, `look through window` — the natural command for a window, and the walkthrough's own step used to reveal the Bio Lab — shows the **room description** instead of the view *through* the window. The near-synonyms all work correctly. Reproduced on **prod** (release `2.0.2`, engine `68f90e8`), narrator OFF:

```
> look through window
Bio Lock East
This is the second half of the sterilization chamber leading from the main lab to the Bio Lab. The door
to the east, leading to the Bio Lab, has a window. The bio lock continues to the west.
There is a multiple purpose robot here.
        ^ the ROOM — no Bio Lab view, no mutants, no card

> look in window
You can see a large laboratory, dimly illuminated. A blue glow comes from a crack in the northern wall of
the lab. Shadowy, ominous shapes move about within the room. On the floor, just inside the door, you can
see a magnetic-striped card.

> look at window     → (same correct Bio Lab view)
> examine window     → (same correct Bio Lab view)
```

So `examine window`, `look at window`, and `look in window` all reveal the Bio Lab (the mutants and the magnetic-striped miniaturization card the player needs to see before the Floyd sacrifice), but `look through window` — the most natural phrasing — does not.

## Root cause

`Planetfall/Location/Lawanda/Lab/BioLockEast.cs:23-30` **explicitly** tries to support "look through window" (note the `OriginalInput.Contains("through")` check), but the through-branch is gated on `action.Match(..., ["window"])`:

```csharp
if (
    (action.Match(new[] { "examine" }.Concat(Verbs.LookVerbs).ToArray(), ["window"])
        && action.OriginalInput != null && action.OriginalInput.Contains("through")) ||   // through-branch
    action.Match(Verbs.ExamineVerbs, ["window"]))                                          // examine/look-at/look-in
    return new PositiveInteractionResult("You can see a large laboratory ... magnetic-striped card. ");

return await base.RespondToSimpleInteraction(...);   // falls through to a bare look -> room
```

[...]

PR #526, opened 2026-07-26
Fix #517: port the SPECK object so "shoot speck with laser" works

### LINKED ISSUE #517: Planetfall: the SPECK object was never ported — `shoot speck with laser` (the canonical solve, and the walkthrough tests' own command) fails in prod with "You don't see any speck here."
## Bug

The microrelay puzzle's target object — the **speck** — does not exist in the port. The relay's description tells the player, in detail, that a speck is the thing wedged in the contact point, and then every command that names it is rejected.

Reproduced on **prod** (`https://6kvs9n5pj4.execute-api.us-east-1.amazonaws.com/Prod/Planetfall`), narrator OFF (`quiet`), engine SHA `42fbf0d` (current `origin/main`):

```
[Strip Near Relay, laser in hand]

> examine relay
This is a vacuum-sealed microrelay, encased in red translucent plastic. Within, you can see
that some sort of speck or impurity has wedged itself into the contact point of the relay,
preventing it from closing. The speck, presumably of microscopic size, resembles a blue
boulder to you in your current size.

> shoot speck with laser
You don't see any speck here.            <-- the canonical original solution

> shoot laser at speck
You don't see any speck here.

> examine speck
This action or command has no effect on the game.      (engine has no handler)

> examine blue boulder
This action or command has no effect on the game.

> examine impurity
This action or command has no effect on the game.

> shoot relay with laser
The laser emits a narrow blue beam of light which slices through the red plastic covering of
the relay like a hot knife through butter. Air rushes into the relay, which collapses into a
heap of plastic shards.                  <-- only the RELAY resolves
```

So the game describes a speck, an impurity, and a blue boulder, and recognises none of those three words. The puzzle is reachable only by naming the **relay** — which is the one framing that reads to a player as "destroy the relay", and which on any non-red setting does exactly that (and silently makes the game unwinnable).

## This is green in CI — the fixture hides it

The walkthrough tests drive this puzzle with **exactly the command that fails in prod**:

`Planetfall.Tests/Walkthrough/WalkthroughTestOne.cs:348,350` (and identically in `WalkthroughByLocationName.cs:361,363` and `WalkthroughDontFixAnything.cs:323,324`):

```csharp
[TestCase("shoot speck with laser", null, "The speck is hit by the beam! It sizzles a little, but isn't destroyed yet.")]
[TestCase("shoot speck with laser", null, "The beam hits the speck again! This time, it vaporizes into a fine cloud of ash. ...")]
```

They pass because the deterministic `TestParser` is handed a mapping that **rewrites the noun** before the engine ever sees it — `UnitTests/IntentMappings/base-mappings.json:320-321`:

```json
{ "inputs": ["shoot relay with laser", "shoot speck with laser"], "verb": "shoot", "nounOne": "relay", "nounTwo": "laser", "preposition": "with", "originalInput": "shoot relay with laser" },
{ "inputs": ["shoot laser at relay", "shoot laser at speck"], "verb": "shoot", "nounOne": "laser", "nounTwo": "relay", "preposition": "at", "originalInput": "shoot laser at relay" },
```

`speck` is mapped to `nounOne: "relay"` in the fixture, so the missing object is invisible to the test suite. In prod the real parser passes `speck` through, `GetItemInScope("speck", …)` returns null, and the command dies. This is the same fixture-shaped blind spot as #423, and precisely what #427 is meant to catch.

## Root cause

There is **no `Speck` item class anywhere in the port** — `grep -rn "speck" Planetfall/ --include=*.cs` returns only string literals inside the relay description and the laser helper's response text. The puzzle was modelled on the relay instead.

`Planetfall/Item/Kalamontee/Mech/Laser.cs:182-205`:

```csharp
private InteractionResult ShootLaserAt(string targetNoun, IContext context)
{
    // Check target scope before firing (don't waste a charge on invalid target)
    var target = Repository.GetItemInScope(targetNoun, context);
    if (target == null)
        return new PositiveInteractionResult($"You don't see any {targetNoun} here. ");
    ...
    // Special response for shooting the Relay
    if (target is Relay relay)
    {
        return LaserSpeckHelper.ShootRelay(relay, context, BeamDescription, Setting, Chooser);
    }
```

The speck's entire state lives on `Relay` as flags (`Planetfall/Item/Computer/Relay.cs:8-22`: `SpeckHit`, `SpeckDestroyed`, `RelayDestroyed`), and the dispatch keys on `target is Relay`. Nothing can resolve the noun `speck`, so the guard on line 186 rejects it before any of that runs.

`SPECK` is a first-class object in the relay, with three synonyms and an adjective — `planetfall-source/comptwo.zil:2562-2567`:

```
<OBJECT SPECK
	(IN RELAY)
	(DESC "speck")
	(SYNONYM SPECK BOULDER IMPURITY)
	(ADJECTIVE BLUE)
	(FLAGS NDESCBIT)>
```

and the laser's `SHOOT` handler dispatches on the **speck** as the indirect object — `comptwo.zil:2712-2714`:

```
	       <COND (<EQUAL? ,PRSI ,SPECK>
		      <SHOOT-SPECK>
		      <RTRUE>)
		     (<EQUAL? ,PRSI ,MICROBE>
		      <SHOOT-MICROBE>
		      <RTRUE>)
```

Note especially: in the original, **both** outcomes live inside `SHOOT-SPECK` (`comptwo.zil:2842-2870`) — the red-setting hit/miss/destroy path, *and* the wrong-setting path that wrecks the relay:

```
<ROUTINE SHOOT-SPECK ()
	<COND (<EQUAL? ,LASER-SETTING 1>
	       ... <SETG SPECK-HIT T> / <REMOVE ,SPECK> <SETG SCORE <+ ,SCORE 8>> ...)
	      (T
	       <REMOVE ,RELAY>
	       <TELL "A thin "> <BEAM-COLOR>
	       <TELL " beam shoots from the laser and slices through the red plastic covering of ...
```

So the original has **no "shoot the relay" path at all** — the player always aims at the speck, and the relay's destruction is a *consequence* of the wrong dial setting. The port inverted the model: the target became the relay and the speck ceased to exist as an object.

## Impact

**Severity: high.** The computer-repair sequence is a required endgame puzzle (it opens the cryo-elevator and awards 8 points), and the canonical command for it — the one in every published Planetfall walkthrough, and the one this repo's own walkthrough tests use — returns a flat refusal in production. It is not a hard block, because `shoot relay with laser` reaches the same code, so the severity is "blocked with an awkward, undiscoverable workaround" rather than critical. But:

- A player who reads the relay description is told the target is the speck, then told the speck isn't there. That contradiction is the entire failure mode.
- The only phrasing that *does* work is the one that most sounds like it will break the puzzle, and with the dial anywhere but "1" it does — destroying the relay and making the game unwinnable with no warning.
- Because it is masked by the fixture, no amount of green CI will surface it; it will keep shipping.

## Suggested fix

Port `SPECK` as a real item and make it the shoot target, matching the ZIL:

1. Add a `Speck : ItemBase, ICanBeExamined` with `NounsForMatching => ["speck", "boulder", "blue boulder", "impurity"]`, held inside `Relay` (it is `NDESCBIT`, so it must not appear in room/container listings), and removed when destroyed.
2. In `Laser.ShootLaserAt`, dispatch on `target is Speck` into the existing `LaserSpeckHelper` logic, keeping `target is Relay` working as a lenient alias so nothing that currently works regresses.
3. Move the existing `SpeckHit` / `SpeckDestroyed` flags onto the new object (or leave them on `Relay` and just re-point the dispatch — the smaller change, and fine if you would rather not move persisted state; note both are `[UsedImplicitly]` and serialized, so moving them needs a save-compat thought).

**Also fix the fixture, or this stays invisible.** `base-mappings.json:320-321` should map `shoot speck with laser` to `nounOne: "speck"` (its honest parse) rather than rewriting it to `relay`. That single change turns the existing walkthrough test cases into real coverage — and they will go red until the `Speck` object exists, which is exactly the signal you want.

## TDD plan

Red-before / green-after, in `Planetfall.Tests`:

1. **Fix the fixture first** (`base-mappings.json:320-321` → `nounOne: "speck"`). The three existing walkthrough test cases that already type `shoot speck with laser` should now **fail** — that is the failing-first step, and it proves the fixture was the mask rather than the handler being fine.
2. Add a focused test at Strip Near Relay: laser held, red battery, dial on 1, `shoot speck with laser` → asserts the "The speck is hit by the beam!" response; a second shot asserts the vaporize text, `+8` score, and that the cryo-elevator door opened.
3. Add the noun-coverage cases the ZIL synonyms require: `shoot boulder with laser`, `shoot impurity with laser`, and `shoot laser at speck` (the `PRSI` phrasing) all reach the same handler; `examine speck` returns a description rather than falling through to the narrator.
4. **Pin what must not regress:** `shoot relay with laser` on a non-red setting still destroys the relay with the existing text, and the walkthroughs still complete end-to-end (`dotnet test Planetfall.Tests`).

## Affected files

- `Planetfall/Item/Computer/Speck.cs` (new), `Planetfall/Item/Computer/Relay.cs`, `Planetfall/Item/Kalamontee/Mech/Laser.cs`, `Planetfall/Item/Kalamontee/Mech/LaserSpeckHelper.cs`
- `UnitTests/IntentMappings/base-mappings.json` (the mask)
- `Planetfall.Tests/Walkthrough/*` (existing cases become real coverage)

Related: #423 (same fixture-hides-prod-gap shape), #427 (the tooling that would catch this class), #228 (missing synonym vocabulary — this is the stronger version: a missing *object*, not a missing synonym).

---

Found via the AdventureBreaker harness (black-box prod + white-box/ZIL confirmation). `AB-095`

arsindelve:
Fixes #517.

## The bug

The microrelay puzzle's target object — the **speck** — was never ported. The relay's description tells the player, in detail, that a speck / impurity / blue boulder is what has wedged itself into the contact point, and then the engine recognised none of those three words. `Repository.GetItemInScope("speck", …)` returned null, and `Laser.ShootLaserAt` bailed out on its scope guard with **"You don't see any speck here."**

The only route through the puzzle was to name the **relay** — which reads to a player as "destroy the relay", and which on any non-red dial setting does exactly that and silently makes the game unwinnable.

## Why CI never caught it

The three walkthroughs drive this puzzle with `shoot speck with laser` — the command that fails in production. They passed because the deterministic fixture rewrote the noun before the engine ever saw it (`UnitTests/IntentMappings/base-mappings.json`):

```json
{ "inputs": ["shoot relay with laser", "shoot speck with laser"], "nounOne": "relay", … }
```

Same fixture-shaped blind spot as #423.

`SPECK` is a first-class object living inside the relay, with three synonyms and an adjective (`comptwo.zil:2562-2567`):

```
<OBJECT SPECK
	(IN RELAY)
	(DESC "speck")
	(SYNONYM SPECK BOULDER IMPURITY)
	(ADJECTIVE BLUE)
	(FLAGS NDESCBIT)>
```

`LASER-F` dispatches on it as the indirect object — `<COND (<EQUAL? ,PRSI ,SPECK> <SHOOT-SPECK>)>` (`comptwo.zil:2712`) — and `SHOOT-SPECK` removes it on the killing shot, `<REMOVE ,SPECK>` (`comptwo.zil:2852`). The non-red branch does `<REMOVE ,RELAY>` (`comptwo.zil:2867`), which takes the speck it contained out of scope with it.

## The fix

- **New `Planetfall/Item/Computer/Speck.cs`** — `Speck : ItemBase, ICanBeExamined`, nouns `speck` / `boulder` / `impurity`, each also accepted with the `blue` adjective. Seeded into `StripNearRelay` alongside the relay. It is NDESC (`ItemBase.GenericDescription` is empty), so it never shows up in the room listing — same as the relay.
- **`Laser.ShootLaserAt`** now dispatches on `target is Speck or Relay`. `Relay` is kept as a lenient alias: the relay's description makes `shoot relay with laser` a natural thing to type, and the port has always accepted it.
- **`LaserSpeckHelper`** takes the speck out of play when it is vaporized, and when a non-red beam shatters the relay containing it — so naming it afterwards fails scope resolution instead of resolving to something that isn't there.
- **The fixture no longer rewrites the noun.** `shoot speck with laser` now maps to `nounOne: "speck"`, so all three walkthroughs exercise the real command and this gap cannot hide from CI again.

The speck's *state* (`SpeckHit`, `SpeckDestroyed`, `MarksmanshipCounter`) deliberately stays on `Relay`, so existing save games and every consumer of `Relay.SpeckDestroyed` (hints, Station 384, the microbe) are unaffected.

### One deliberate divergence

The original gives `SPECK` no `ACTION` routine, so examining it falls through to the default "nothing special" message. In this engine that message is built from the item's **longest** noun, yielding "There is nothing special about the blue impurity." — which reads badly immediately after the relay has spent three sentences describing the speck. `examine speck` therefore reuses the relay's own wording instead. Flagged here so it's a review decision, not an accident.

## TDD

Red first: with the `Speck` type present but unwired, the new tests failed with exactly the production message from the issue —

```
Failed ShootLaserAtSpeck_WithRedLaser_FirstHit_SetsSpeckHit
   Expected response "You don't see any speck here.
Failed ShootSpeck_BySynonym_WithRedLaser_HitsTheSpeck("boulder")
   Expected response "You don't see any boulder here.
Failed ShootSpeck_BySynonym_WithRedLaser_HitsTheSpeck("impurity")
   Expected response "You don't see any impurity here.
Failed Speck_IsPresentInTheRelayRoom
   Expected speck.CurrentLocation to be StripNearRelay … but found <null>.
```

10 failed / 40 passed before the fix; 50 passed after.

New tests in `Planetfall.Tests/RelayTests.cs` (`#region Speck Tests`): presence and non-listing in the room, `examine speck` and both synonyms, `shoot speck with laser` and `shoot laser at speck` first/second hit, synonym shots, the non-red destroy-the-relay path, and both out-of-scope-after-destruction paths. All deterministic (`IRandomChooser` mocked, real `Repository` with `Reset()`), all in the normal test command.

## Verification

Every suite that touches the changed code, run separately — all green:

| Suite | Result |
|---|---|
| `Planetfall.Tests` | 3283 passed, 0 failed |
| `UnitTests` | 1130 passed, 0 failed |
| `ZorkOne.Tests` | 1782 passed, 0 failed |
| `EscapeRoom.Tests` | 9 passed, 0 failed |
| `Stationfall.Tests` | 4 passed, 0 failed |
| `Console.Tests` | 16 passed, 0 failed |

https://claude.ai/code/session_01JQR4n8dnPcVgbN19pMw7xC

---
_Generated by [Claude Code](https://claude.ai/code/session_01JQR4n8dnPcVgbN19pMw7xC)_

### DIFF
--- a/Planetfall.Tests/RelayTests.cs
+++ b/Planetfall.Tests/RelayTests.cs
@@ -427,6 +427,184 @@ public class RelayTests : EngineTestsBase
 
     #endregion
 
+    #region Speck Tests
+
+    // Issue #517: the SPECK is a first-class object in the original (comptwo.zil:2562-2567,
+    // SYNONYM SPECK BOULDER IMPURITY / ADJECTIVE BLUE) and the laser's SHOOT handler dispatches on
+    // it (comptwo.zil:2712). The port modelled the puzzle on the relay alone, so the canonical
+    // solution "shoot speck with laser" died on "You don't see any speck here." in production —
+    // hidden from CI only because the test fixture rewrote the noun to "relay".
+
+    [Test]
+    public void Speck_IsPresentInTheRelayRoom()
+    {
+        GetTarget();
+        var location = StartHere<StripNearRelay>();
+        var speck = GetItem<Speck>();
+
+        speck.CurrentLocation.Should().Be(location);
+        location.Items.Should().Contain(speck);
+    }
+
+    [Test]
+    public async Task Speck_IsNotListedInTheRoomDescription()
+    {
+        var target = GetTarget();
+        StartHere<StripNearRelay>();
+
+        var response = await target.GetResponse("look");
+
+        // NDESCBIT in the original — the speck is only visible through the relay's description.
+        response.Should().NotContain("There is a speck");
+    }
+
+    [Test]
+    public async Task ExamineSpeck_DescribesTheSpeck()
+    {
+        var target = GetTarget();
+        StartHere<StripNearRelay>();
+
+        var response = await target.GetResponse("examine speck");
+
+        response.Should().Contain("blue boulder");
+    }
+
+    [TestCase("boulder")]
+    [TestCase("impurity")]
+    public async Task ExamineSpeck_BySynonym_DescribesTheSpeck(string noun)
+    {
+        var target = GetTarget();
+        StartHere<StripNearRelay>();
+
+        var response = await target.GetResponse($"examine {noun}");
+
+        response.Should().Contain("blue boulder");
+    }
+
+    [Test]
+    public async Task ShootSpeck_WithRedLaser_FirstHit_SetsSpeckHit()
+    {
+        var target = GetTarget();
+        StartHere<StripNearRelay>();
+        Take<Laser>();
+        GetItem<OldBattery>().ChargesRemaining = 10;
+        GetItem<Laser>().Setting = 1; // Red
+        SetupMockRandomChooser(shouldHit: true);
+        var relay = GetItem<Relay>();
+
+        var response = await target.GetResponse("shoot speck with laser");
+
+        response.Should().NotContain("You don't see any speck here");
+        response.Should().Contain("The speck is hit by the beam!");
+        relay.SpeckHit.Should().BeTrue();
+        relay.SpeckDestroyed.Should().BeFalse();
+    }
+
+    [Test]
+    public async Task ShootLaserAtSpeck_WithRedLaser_FirstHit_SetsSpeckHit()
+    {
+        var target = GetTarget();
+        StartHere<StripNearRelay>();
+        Take<Laser>();
+        GetItem<OldBattery>().ChargesRemaining = 10;
+        GetItem<Laser>().Setting = 1; // Red
+        SetupMockRandomChooser(shouldHit: true);
+        var relay = GetItem<Relay>();
+
+        var response = await target.GetResponse("shoot laser at speck");
+
+        response.Should().Contain("The speck is hit by the beam!");
+        relay.SpeckHit.Should().BeTrue();
+    }
+
+    [TestCase("boulder")]
+    [TestCase("impurity")]
+    public async Task ShootSpeck_BySynonym_WithRedLaser_HitsTheSpeck(string noun)
+    {
+        var target = GetTarget();
+        StartHere<StripNearRelay>();
+        Take<Laser>();
+        GetItem<OldBattery>().ChargesRemaining = 10;
+        GetItem<Laser>().Setting = 1; // Red
+        SetupMockRandomChooser(shouldHit: true);
+
+        var response = await target.GetResponse($"shoot {noun} with laser");
+
+        response.Should().Contain("The speck is hit by the beam!");
+    }
+
+    [Test]
+    public async Task ShootSpeck_WithRedLaser_SecondHit_DestroysSpeck()
+    {
+        var target = GetTarget();
+        StartHere<StripNearRelay>();
+        Take<Laser>();
+        GetItem<OldBattery>().ChargesRemaining = 10;
+        GetItem<Laser>().Setting = 1; // Red
+        SetupMockRandomChooser(shouldHit: true);
+        var relay = GetItem<Relay>();
+        relay.SpeckHit = true; // Already hit once
+
+        var response = await target.GetResponse("shoot speck with laser");
+
+        response.Should().Contain("The beam hits the speck again!");
+        response.Should().Contain("vaporizes into a fine cloud of ash");
+        relay.SpeckDestroyed.Should().BeTrue();
+    }
+
+    [Test]
+    public async Task ShootSpeck_WithNonRedLaser_DestroysTheRelay()
+    {
+        var target = GetTarget();
+        StartHere<StripNearRelay>();
+        Take<Laser>();
+        GetItem<OldBattery>().ChargesRemaining = 10;
+        GetItem<Laser>().Setting = 3; // Yellow (non-red)
+
+        var response = await target.GetResponse("shoot speck with laser");
+
+        response.Should().Contain("slices through the red plastic");
+        GetItem<Relay>().RelayDestroyed.Should().BeTrue();
+    }
+
+    [Test]
+    public async Task ShootSpeck_AfterSpeckDestroyed_SpeckIsGoneFromScope()
+    {
+        var target = GetTarget();
+        StartHere<StripNearRelay>();
+        Take<Laser>();
+        GetItem<OldBattery>().ChargesRemaining = 10;
+        GetItem<Laser>().Setting = 1; // Red
+        SetupMockRandomChooser(shouldHit: true);
+        GetItem<Relay>().SpeckHit = true; // Already hit once — this shot vaporizes it
+
+        await target.GetResponse("shoot speck with laser");
+        var response = await target.GetResponse("shoot speck with laser");
+
+        // <REMOVE ,SPECK> (comptwo.zil:2852) — the speck no longer exists to be shot at.
+        response.Should().Contain("You don't see any speck here");
+        GetItem<Speck>().CurrentLocation.Should().BeNull();
+    }
+
+    [Test]
+    public async Task ShootSpeck_AfterRelayDestroyed_SpeckIsGoneFromScope()
+    {
+        var target = GetTarget();
+        StartHere<StripNearRelay>();
+        Take<Laser>();
+        GetItem<OldBattery>().ChargesRemaining = 10;
+        GetItem<Laser>().Setting = 4; // Green (non-red) — shatters the relay
+
+        await target.GetResponse("shoot speck with laser");
+        var response = await target.GetResponse("shoot speck with laser");
+
+        // <REMOVE ,RELAY> (comptwo.zil:2867) takes the speck it contained out of scope with it.
+        response.Should().Contain("You don't see any speck here");
+        GetItem<Speck>().CurrentLocation.Should().BeNull();
+    }
+
+    #endregion
+
     #region Timer Tests
 
     [Test]
--- a/Planetfall/Item/Computer/Speck.cs
+++ b/Planetfall/Item/Computer/Speck.cs
@@ -1,0 +1,47 @@
+namespace Planetfall.Item.Computer;
+
+/// <summary>
+/// The impurity wedged into the microrelay's contact point — the actual target of the endgame
+/// laser puzzle.
+///
+/// Issue #517: this object was never ported. The relay's description tells the player, in detail,
+/// that a speck is what has jammed the relay, and the canonical solution (both the original
+/// walkthroughs' and this repo's own walkthrough tests') is "shoot speck with laser" — but with no
+/// Speck in the Repository, <c>GetItemInScope("speck")</c> returned null and the laser answered
+/// "You don't see any speck here." The only way through the puzzle was to name the *relay*, which
+/// on any non-red dial setting shatters it and silently makes the game unwinnable.
+///
+/// In the original the speck is a first-class object living inside the relay
+/// (comptwo.zil:2562-2567) with three synonyms and an adjective, and the laser's SHOOT handler
+/// dispatches on it (comptwo.zil:2712 -> SHOOT-SPECK).
+///
+/// The speck's *state* deliberately stays on <see cref="Relay" /> (SpeckHit / SpeckDestroyed /
+/// MarksmanshipCounter) so existing save games and the several consumers of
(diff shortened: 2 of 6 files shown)


arsindelve:
Review follow-up — all three points triaged, one applied in `a171a46`, two settled as deliberate calls.

### Applied: use the engine's own destroy primitive

`Speck.RemoveFromPlay()` was byte-for-byte `Repository.DestroyItem<T>()` (`GameEngine/Repository.cs:467`). Deleted in favor of the shared helper. No behavior change; suites re-run green (Planetfall 3283, UnitTests 1130, ZorkOne 1782).

### Keeping the `Relay` alias — deliberate

The review is right about the original: `LASER-F` dispatches `SHOOT-SPECK` only on `PRSI == SPECK` (comptwo.zil:2712), and `PRSI == RELAY` falls through to the generic "grows a bit warm, but nothing else happens" branch. Dropping the alias would also remove a real footgun — the dial defaults to 5, so `shoot relay with laser` shatters it and silently makes the game unwinnable.

But #517's own Proposed Solution asks for it directly:

> 3. Keep `target is Relay` working as a lenient alias for backward compatibility

Keeping it, per the issue. Worth noting the alias is *only* a leniency now — the puzzle is fully solvable through the speck, which is what #517 was actually about. If the unwinnable-state footgun is worth closing on its own terms, that's a clean follow-up: dispatch only on `Speck` and retarget the ~10 existing `ShootRelay_*` cases.

### Legacy in-flight sessions still won't see the speck — known, not fixed here

Confirmed independently. `Repository.Restore` (`GameEngine/Repository.cs:348`) swaps `_allItems` / `_allLocations` wholesale and never re-runs `Init()`, so a session saved by the pre-fix build restores a `StripNearRelay` whose `Items` has no `Speck`. I reproduced it by round-tripping a save with the `Speck` entries stripped from the JSON (both the location's `Items` array and the `AllItems` key — the shape a pre-#517 save actually has):

```
SPECK STILL IN LEGACY SAVE: False
RESPONSE: You don't see any speck here.
```

Not fixed in this PR, for two reasons:

1. **It needs machinery that doesn't exist.** `IInfocomGame.Init` runs only at construction (`GameEngine.cs:100,144,773`), never on restore, and `RestoreGame`'s existing `RequestSequence` net is engine-level state — a Planetfall-specific speck migration doesn't belong there. The alternative, `[OnDeserialized]` on the location, runs mid-graph and would touch the pre-restore Repository. Either way that's new engine machinery inside a bug-fix PR.
2. **It isn't specific to this change.** Every item ever added to a room in this repo has the same gap. A general post-restore migration hook would fix all of them at once; a bespoke speck heal fixes exactly one.

Scope of impact: only sessions already parked on the silicon strip. A new game gets the fix, and a heal condition is available whenever it's wanted — speck absent while both `SpeckDestroyed` and `RelayDestroyed` are false is unambiguous, since a legitimately destroyed speck always leaves one of those flags set.

---

arsindelve:
Correction to my previous comment: I claimed dropping the `Relay` alias would "remove a real footgun — the dial defaults to 5, so `shoot relay with laser` shatters it and silently makes the game unwinnable." The unwinnable state is real, but that rationale is wrong, and it was the review's main argument for dropping the alias.

In the original, the destructive branch lives *inside* `SHOOT-SPECK`, which is dispatched on `PRSI == SPECK` (comptwo.zil:2842-2872):

```
<ROUTINE SHOOT-SPECK ()
	<COND (<EQUAL? ,LASER-SETTING 1>
	       ... hit/miss logic on the speck ...)
	      (T
	       <REMOVE ,RELAY>
	       <TELL "... slices through the red plastic covering of the relay like a
	              hot knife through butter. Air rushes into the relay, which
	              collapses into a heap of plastic shards." CR>)>>
```

So the relay is shattered by shooting the **speck** on any non-red setting — and the dial starts wrong, `<GLOBAL LASER-SETTING 5>` (comptwo.zil:2631), which this port matches. Since `SHOOT-SPECK`'s success branch is the only thing that sets `COMPUTER-FIXED` (and with it `<FSET ,CRYO-ELEVATOR-DOOR ,OPENBIT>`), that does end the run silently. Authentic Infocom cruelty, faithfully ported: `ShootSpeck_WithNonRedLaser_DestroysTheRelay` covers it.

Shooting the relay *by name* is the inert case in the original — `RELAY` isn't in the burnable list, so it falls through to "which strikes the micro-relay. The micro-relay grows a bit warm, but nothing else happens."

Consequence for the alias decision: dropping `target is Relay` would **not** close the footgun. `shoot speck with laser` at the default setting is the canonical command and stays fully destructive either way; dropping the alias would only make `shoot relay` inert, moving the trap back where Infocom put it rather than removing it. The alias therefore costs nothing the authentic path doesn't already cost, which makes keeping it (as #517 asked) the cheaper call. The only thing still on the table is the pure-faithfulness argument: in the original, `shoot relay` warms the relay and nothing more.

---

Issue #545, opened 2026-08-20
Floyd's corpse chats and squeals: post-sacrifice interactions don't consult HasDied


arsindelve:
## Summary

Immediately after the Bio-Lab sacrifice — the game's emotional climax, which itself plays out perfectly (volunteer speech, knock choreography, full death scene with the Ballad of the Starcrossed Miner, +2 points, body and card placed) — Floyd's corpse still responds to the player as if alive. Three manifestations, one root cause: post-mortem interaction paths key off `IsOn` (which `EndSequence` deliberately leaves `true` on the corpse) and never consult `HasDied`.

All reproduced live in prod (2.0.7), one turn after the death scene, while `examine floyd` correctly refuses out of grief ("a tremendous sense of loss overcomes you").

## Manifestations

**1. The corpse chats (worst).**
```
> floyd, are you okay
"Floyd says he is okay now that you are here."
```
`Floyd.OnBeingTalkedTo` (`Planetfall/Item/Kalamontee/Mech/FloydPart/Floyd.cs:169-193`) gates only on `!IsOn`. The corpse's `IsOn` is deliberately left `true` by `EndSequence` (documented at `BioLockStateMachineManager.cs:63-68`), so `ConversationHandler` routes conversation with the body straight to the chat lambda, which replies cheerfully.

**2. The corpse squeals and scoots away.**
```
> take floyd
You manage to lift Floyd a few inches off the ground, but he is too heavy and you
drop him suddenly. Floyd gives a surprised squeal and moves a respectable distance away.
```
`CannotBeTakenDescription => IsOn ? FloydConstants.TakeFloyd : null` (`Floyd.cs:123-125`) serves the living-Floyd refusal for the same reason.

**3. (Softer, trapped-death branch.)** If Floyd dies trapped in the lab (door never reopened), one turn after "Then, silence. Floyd is dead." the narrator answers questions about him with jokes: "Floyd must be off on his own little adventure" / "off on his own space adventure elsewhere." Related to the absent-NPC handling in AB-019, but the death context makes it a tonal problem of its own.

## Control case proving the pattern

`turn off floyd` on the corpse is handled beautifully: *"I'm afraid that Floyd has already been turned off, permanently, and gone to that great robot shop in the sky."* The #493 sweep (corpse cheerfully asking to open the door) fixed that class of bug for the actor/turn path — these are the remaining interaction paths it missed.

## Also verified while here (no action needed)

[...]

PR #556, opened 2026-08-20
Close two gone-for-good conversation gaps left by #545's second review

arsindelve:
A second `/code-review` round ran on [#548](https://github.com/arsindelve/ZorkAI/pull/548) at 17:09, six minutes before it merged at 17:15. Nine findings, **none actioned**. This PR takes the two that are live player-facing regressions; the rest are triaged at the bottom.

Both reproduce the same way: a **blank response** — the utterance leaking out of the conversation system entirely, which is exactly the failure the gone-for-good route was created to stop.

## 1. Nameless speech to a corpse still blanks offline

`TryRouteNamelessSpeech` never got the `IsGoneForGood` gate that `CheckForConversation`'s *named* branch got. Standing over Floyd's body with `NoGeneratedResponses` set (a per-request flag the Planetfall API exposes) and saying `hello`:

```
> say hello
                       <-- blank
```

It bails at the generation kill-switch before ever reaching `OnBeingTalkedTo`. A gone-for-good talker's reply is a compile-time constant and owes generation nothing on this branch either.

## 2. Loose address phrasings stopped reaching the corpse

`RespondForGoneForGoodTalker` replaced a two-part detector — the `ParseConversation` classifier **OR** the deterministic strip — with `IsGenuineDirectAddress` alone.

That compared against the wrong baseline. `IsGenuineDirectAddress` *is* broader than `TryStripDirectAddress`, but it is **not** broader than the pipeline actually replaced. Phrasings only the classifier recognized — `could you let floyd know to wait for me`, which leads with neither a name nor an address lead-in — stopped reaching Floyd and leaked to the narrator, who then improvises about a robot the player watched die. The method's own XML doc asserted the opposite ("strictly more phrasings than before"); **corrected here**.

The fix restores the classifier as a backstop, mirroring `FindAddressedAbsentCharacter` exactly — including its rule of skipping the classifier when generation is disabled, so the guard stays deterministic offline.

**The round-trip saving is preserved.** `&&` short-circuits, so the classifier is never consulted for a phrasing `IsGenuineDirectAddress` already recognizes — the overwhelming majority. The existing `TalkToFloyd_DeadAndPresent_DoesNotCallTheConversationClassifier` passes unchanged, which is the proof.

## Tests

Three new deterministic tests in `FloydTests`, red before the fix and red for the right reason (both returned `""`):

- `SpeakingNamelesslyToDeadFloyd_MournsEvenWithGenerationDisabled`
- `AddressingDeadFloydWithUnusualPhrasing_DefersToClassifier_AndMourns`
- `AddressingDeadFloydWithUnusualPhrasing_GenerationDisabled_DoesNotConsultClassifier` — pins the offline determinism, so the backstop can't quietly become an online-only dependency.

All suites green: Planetfall 3960, UnitTests 1180, ZorkOne 1782, EscapeRoom 9, Stationfall 62, Console 16.

## The rest of that review round, triaged

I verified all nine against `main`. All are real; these six are not in this PR:

| Finding | Kind |
|---|---|
| `SleepEngine.cs:273` — corpse teleports to you nightly | **Fixed in [#555](https://github.com/arsindelve/ZorkAI/pull/555)** |
| `BoothBase.cs:57` — booth gates on presence with no liveness test | Latent; was reachable only via the SleepEngine bug, worth fixing on its own merits |
| `Floyd.Act():301` — `HasDied` return sits above the block clearing `PendingFloydActionCommentPrompt`, so a prompt queued the turn Floyd dies is burned unspoken and occupies the slot permanently in saved state | Real, narrow |
| `WalkthroughBioLock.cs:41` — fixture reset covers `HasDied` but not `IsAwayOnScriptedSequence`, `IsOffWandering`, `ItemBeingHeld`, `HasRevealedLowerElevatorCard` | Test fragility |
| `FloydConstants.cs:66` — header says "The three constants below"; there are five | Cosmetic |
| `Docs/hints/planetfall/02-state-audit.md:53` — still names the renamed `IsHereAndIsOn` | Stale doc |

(Round 1 of that review — a separate 10 findings — was worked to completion and shipped in #548; I verified each against `main`.)

### DIFF
--- a/GameEngine/ConversationHandler.cs
+++ b/GameEngine/ConversationHandler.cs
@@ -80,19 +80,27 @@ public class ConversationHandler(
     /// <c>ParseConversation</c> rewriter the living path uses - that call is an AWS Lambda round-trip
     /// whose only product is a command the character will never act on.
     ///
-    /// Address detection therefore falls to <see cref="IsGenuineDirectAddress"/>, the same
-    /// deterministic test the absent path relies on. That is not a downgrade from the living path's
-    /// backstop: <see cref="TryStripDirectAddress"/> recognizes only the leading-name forms and leaves
-    /// imperative lead-ins ("ask floyd about the card") to the classifier, whereas
-    /// IsGenuineDirectAddress covers both - so a corpse is addressable by strictly more phrasings than
-    /// before, and deterministically, including in NoGeneratedResponses mode.
+    /// Address detection is two-tier, mirroring the absent path exactly: the deterministic
+    /// <see cref="IsGenuineDirectAddress"/> first - which covers leading names AND imperative
+    /// lead-ins, so the common phrasings answer offline and without any round-trip - and the
+    /// classifier only as a backstop for what it misses.
+    ///
+    /// The backstop is not optional. An earlier version of this method used IsGenuineDirectAddress
+    /// alone, on the reasoning that it is strictly broader than <see cref="TryStripDirectAddress"/>.
+    /// That compared against the wrong thing: the pipeline actually replaced was the classifier OR
+    /// the strip, and looser phrasings the classifier alone recognized ("could you let floyd know
+    /// ...", which leads with neither a name nor a lead-in) stopped reaching the character and leaked
+    /// to the narrator - the exact outcome this route exists to prevent.
     ///
     /// Returns null when the input merely NAMES the character rather than addressing them ("examine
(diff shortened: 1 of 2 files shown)


arsindelve:
Folded into #554 at the owner's request — same commit, cherry-picked onto that branch, so nothing is lost. Review it there.

Closing rather than merging so the work lands as a single PR. Branch `claude/gone-for-good-conversation-gaps` is left in place but is now redundant.

